PR #3, opened 2023-07-04
Add Ruby support

tradiff:
Ruby doesn't use brackets for logic blocks, so there's not a lot we can do here. But I did what I can. I mostly matched what VS-Code does for Ruby files. The only difference I'm aware of is I also included `do_block` which I think is not TOO bad, and could be useful for deeply nested specs like the example provided.

### DIFF
--- a/queries/ruby/rainbow-delimiters.scm
+++ b/queries/ruby/rainbow-delimiters.scm
@@ -1,0 +1,15 @@
+(block
+  "{" @opening
+  "}" @closing) @container
+
+(hash
+  "{" @opening
+  "}" @closing) @container
+
+(array
+  "[" @opening
+  "]" @closing) @container
+
+(parenthesized_statements
+  "(" @opening
+  ")" @closing) @container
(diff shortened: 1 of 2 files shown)


HiPhish:
Looks good, but please rebase on the current `master` first. There is a nasty bug that breaks everything if the user has a custom configuration in the old version.

I am not certain about adding the `do` block pattern to this as well. People might find it too vibrant if entire words are highlighted as well, plus this kind of nesting is usually make obvious enough through indentation. Lua has the same issue and my solution is to have two queries: `rainbow-delimiters` only highlights parenthesis-like characters while `rainbow-blocks` highlights keyword delimiters on top of that as well. That way users can decide how vibrant they want their code to be. You could even add more delimiter pairs like `class`/`end`, `def`/`end` or `if`/`else`/`end` if it makes sense.

I don't know if there is a way of combining queries, so I had to copy-paste definitions from `rainbow-delimiters` into `rainbow-blocks`. If you find a better way please let me know.

tradiff:
@HiPhish I've rebased and remove `do_block`. I'm afraid I don't have the bandwidth right now to implement rainbow-blocks.

HiPhish:
Thank you, looks good. There was no `block` entry in the sample file, so I added it myself.

> I'm afraid I don't have the bandwidth right now to implement rainbow-blocks.

That's fine, don't worry. One query at a time.

PR #5, opened 2023-07-06
Support SystemVerilog and Verilog

mattyoung101:
This PR adds some support for Verilog and SystemVerilog, both of which are covered by the same Treesitter grammar in Neovim.

`begin` and `end` blocks should be coloured correctly, which are the main blocks in both languages. I've included support for some extra things like bit indexes. Unfortunately, the SystemVerilog grammar is not very reliable at the moment, which can sometimes cause issues with highlighting. I was also unable to get nested expressions like `(a+(a+b))` to highlight fully correctly.

### DIFF
--- a/doc/rainbow-delimiters.txt
+++ b/doc/rainbow-delimiters.txt
@@ -295,6 +295,9 @@ addition there are the following extra queries for certain languages:
   - `rainbow-parens` Only parentheses without React tags
 - `tsx`
   - `rainbow-parens` Just Typescript highlighting without React tags
+- `verilog`
+  - `rainbow-blocks` Matches keyword delimiters like `begin` and `end`, in
+    addition to parantheses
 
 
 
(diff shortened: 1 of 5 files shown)


HiPhish:
Thank you, but I don't think keyword delimiters like `begin`/`end` should be part of the default query. I already mentioned this in PR #3, but I prefer `rainbow-delimiters` to be parentheses only and `rainbow-blocks` to be parentheses and keyword-like delimiters. I have already done the same for Lua if you want to see it for reference.

The reason for this is that if we highlight entire words the text can become too vibrant and exhausting to look at. So I prefer the default query to be more conservative. The only exception are languages like HTML where entire words delimiters are the only delimiters, so it doesn't make sense to have any other default query. I should probably write this down in the CONTRIBUTING file.

mattyoung101:
Thanks for the feedback! I will update the PR accordingly.

HiPhish:
Merged, thank you very much.

PR #7, opened 2023-07-10
Color C++ namespaces

MCredstoner2004:
Makes cpp query also match namespace declarations in C++


HiPhish:
Thank you. Nest time please add a small sample of the pattern you are matching to the highlight test file. It does not have to be anything complicated, just enough to match the pattern. For C++ that's `test/highlight/cpp/regular.cpp`. It's not a big deal here because I know enough C++, but I might not be able to do it for other languages.

PR #13, opened 2023-07-17
C# - Implicit array creation

jcapblancq:
Hi there!

Added a missing delimiter highlighting for C#.

Before:
image

After:
image


HiPhish:
Merged, thank you.

HiPhish:
I have tried your configuration and it works for me. Does it work in Lua
if you don't have any configuration at all? Does it work if you `:write`
and then `:edit` the file?

PR #15, opened 2023-07-21
fix(health): remove deprecated API warnings

gh-liu:
Remove the deprecated check health API warnings, see [here](https://neovim.io/doc/user/deprecated.html#_functions).


HiPhish:
Thank you. I rarely follow the current state of development, so PRs for the upcoming release are a great help to me.

PR #22, opened 2023-08-09
feat(sql): add support for sql

liljaylj:
add support for sql

### DIFF
--- a/queries/sql/rainbow-delimiters.scm
+++ b/queries/sql/rainbow-delimiters.scm
@@ -1,0 +1,19 @@
+(term
+    "(" @opening
+    ")" @closing) @container
+
+(subquery
+    "(" @opening
+    ")" @closing) @container
+
+(invocation
+    "(" @opening
+    ")" @closing) @container
+
+(list
+    "(" @opening
+    ")" @closing) @container
+
+(binary_expression
+    "(" @opening
+    ")" @closing) @container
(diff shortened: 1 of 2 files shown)


HiPhish:
Does this actually work for you? I tried some simple queries and this one is giving me problems:

```sql
select *
from products
where (
	Product_Category = 'Fit'
	and  Product_number IN (1234, 1235, 1236, 1237, 1238)
)
or
	(Product_Category in ('Tight', 'Wide') and Product_number = 1324);
```

The first pair of parentheses is blue (level 2) and the second one is red (level 1), but both should be red (level 1).

Screenshot_20230809_155638

I think the problem is that in the tree all parentheses are on the same level, which means the pattern

```
(binary_expression
    "(" @opening
    ")" @closing) @container
```

matches multiple `@opening` and `@closing` captures. This might resolve on its own when neovim/neovim#17099 gets solved in Neovim. But I am just guessing here. What is your opinion, should I just merge it as is and hope for the best or should we wait and see? The question is which is the lesser evil, sometimes broken highlight pairs or no highlight pairs at all?

In either case please write a small test file here as well.

HiPhish:
No, they are still part of *some* `binary_expression`, just not the one you are looking at. Press `a` in the TS Playground to toggle the anonymous nodes and you will see where in the tree the parentheses are. This would not be a problem if there was some `parenthesized_expression` nod that wraps the pair like in some other languages, but here all the parentheses are children of the same `@container` node. This makes the `@opening` and `@closing` capture match multiple nodes each, which causes chaos.

HiPhish:
Looks good. Especially the last pattern, I am quite surprised that it actually works.

PR #24, opened 2023-08-10
fix(php): remove buggy string interpolation curly braces

liljaylj:
image

this snippet gives 6! combinations of opening and closing curly braces (because they all on same level in the tree). therefore it messes up highlight.

i think it would be better to remove this query. especially since there won't be encapsulation inside another encapsulation (i don't think it's even possible)


HiPhish:
I think the pattern is useful, just not for the current state of Neovim. I have commented it out and added a remark to the top of the pattern. We can comment it back in once the issue is resolved.

PR #33, opened 2023-08-26
feat(perl): add support for Perl

delphinus:
This adds support for Perl


HiPhish:
Thank you for your PR. I get a parser error on line 31 of the test file, can you please fix that? I don't know Perl, so I cannot do it myself. And please rename the test file to `test/highlight/perl/regular.pl` to be more consistent with the rest.

delphinus:
@HiPhish sorry for toooo late.

* fixed filenames: `code.pl` → `regular.pl`
* I noticed tree-sitter-perl does not support string interpolation yet. so I commented out the code.

check it!

HiPhish:
Thank you. I have squashed the commits and cleaned up a few details before merging.

HiPhish:
d

PR #35, opened 2023-09-10
feat(fsharp): add F# support

SoxPopuli:
Adds queries for F# files.

Treesitter doesn't have a grammar for F# files by default, but one can be installed from [this repo](https://github.com/Nsidorenco/tree-sitter-fsharp). As far as I am aware, this is the only actively developed custom grammar for F# so the chances of people installing the wrong grammar should be minimal, that being said, I would understand if you were to reject this PR on those grounds.

You should be able to run the test file directly using the dotnet F# repl: `dotnet fsi <path-to-file>`


Danielkonge:
To be able to add the F# queries, they will be need to be updated to the new syntax. 

For the first one you commented out, something like this might work?
```query
(argument_patterns
  "(" @delimiter
  ("(" @delimiter ")" @delimiter)*
  ")" @delimiter @sentinel
) @container
```
(It depends on how the F# parser's tree structure looks, and I haven't checked it.)

HiPhish:
I have gone ahead and updated the queries to fit the parser from nvim-treesitter. I hope there is nothing missing or wrong, this was my best effort guess.

PR #43, opened 2023-10-06
Rainbow blocks fix for Lua (+ other languages with a little work)

Danielkonge:
This is a draft of a fix to the problem where rainbow-delimiters doesn't catch the same group multiple times. E.g., it couldn't catch `elseif` and `else` in Lua before. 

I have time to edit and add improvements if you have any suggestions, or we could add it as another strategy if you don't want it to replace the old `global`? 

I have only tested it in Lua and Rust for now (and before getting rainbow blocks to work better than before in other languages we need to edit the corresponding .scm files). But it works exactly as I would expect so far. 

To use this it is important that each `@container` only has at most one `@closing`, but you can have as many `@opening` or `@intermediate` as you want. 

Screenshot 2023-10-06 at 22 33 51

### DIFF
--- a/lua/rainbow-delimiters/lib.lua
+++ b/lua/rainbow-delimiters/lib.lua
@@ -64,7 +64,7 @@ M.buffers = {}
 ---Fetches the query object for the given language from the settings.
 ---
 ---@param lang string  Name of the language to get the query for
----@return userdata query  The query object
+---@return Query? query  The query object
 function M.get_query(lang)
 	local name = config['query'][lang]
 	local query = get_query(lang, name)
@@ -211,7 +211,7 @@ function M.detach(bufnr)
 	local parser = M.buffers[bufnr].parser
 
 	-- Clear all the namespaces for each language
-	util.for_each_child(parser:lang(), parser, function(_, lang)
+	util.for_each_child(nil, parser:lang(), parser, function(_, lang)
 		M.clear_namespace(bufnr, lang)
 	end)
 	-- Finally release all resources the parser is holding on to
--- a/lua/rainbow-delimiters/strategy/global.lua
+++ b/lua/rainbow-delimiters/strategy/global.lua
@@ -19,7 +19,6 @@ local Stack = require 'rainbow-delimiters.stack'
 local lib   = require 'rainbow-delimiters.lib'
 local util  = require 'rainbow-delimiters.util'
 local log   = require 'rainbow-delimiters.log'
-local ts    = vim.treesitter
 
 
 ---Strategy which highlights the entire buffer.
@@ -28,33 +27,52 @@ local M = {}
 ---Changes are range objects and come in two variants: one with four entries and
 ---one with six entries.  We only want the four-entry variant.  See
 ---`:h TSNode:range()`
+---@param change number[]
+---@return number[]
 local function normalize_change(change)
 	local result
 	if #change == 4 then
 		result = change
-	else
+	elseif #change == 6 then
 		result = {change[1], change[2], change[4], change[5]}
+	else
+		result = {}
 	end
 	return result
 end
 
+---@param bufnr number
+---@param lang string
+---@param matches Stack
+---@param level number
+---@return nil
 local function highlight_matches(bufnr, lang, matches, level)
 	local hlgroup = lib.hlgroup_at(level)
 	for _, match in matches:iter() do
-		for _, opening      in match.opening:iter()      do lib.highlight(bufnr, lang, opening,      hlgroup) end
-		for _, closing      in match.closing:iter()      do lib.highlight(bufnr, lang, closing,      hlgroup) end
-		for _, intermediate in match.intermediate:iter() do lib.highlight(bufnr, lang, intermediate, hlgroup) end
+		for _, delimiter in match.delimiter:iter() do lib.highlight(bufnr, lang, delimiter, hlgroup) end
 		highlight_matches(bufnr, lang, match.children, level + 1)
 	end
+end
+
+
+---Create a new empty match_record
+---@return table
+local function new_match_record()
+	return {
+		delimiter = Stack.new(),
+		children = Stack.new(),
+	}
 end
 
 ---Update highlights for a range. Called every time text is changed.
 ---@param bufnr   number  Buffer number
 ---@param changes table   List of node ranges in which the changes occurred
----@param tree    table   TS tree
+---@param tree    TSTree  TS tree
 ---@param lang    string  Language
+---@return nil
 local function update_range(bufnr, changes, tree, lang)
-	log.debug('Updated range with changes ' .. vim.inspect(changes))
+	log.debug('Updated range with changes %s', vim.inspect(changes))
+
 	if not lib.enabled_for(lang) then return end
 	if vim.fn.pumvisible() ~= 0 or not lang then return end
 
@@ -64,35 +82,59 @@ local function update_range(bufnr, changes, tree, lang)
 	local matches = Stack.new()
 
 	for _, change in ipairs(changes) do
+		-- This is the match record, it lists all the relevant nodes from
+		-- each match.
+		---@type table?
+		local match_record
 		local root_node = tree:root()
 		local start_row, end_row = change[1], change[3] + 1
 		lib.clear_namespace(bufnr, lang, start_row, end_row)
-		for _, match, _ in query:iter_matches(root_node, bufnr, start_row, end_row) do
-			-- This is the match record, it lists all the relevant nodes from
-			-- the match.
-			local match_record = {
-				opening = Stack.new(),
-				closing = Stack.new(),
-				intermediate = Stack.new(),
-				children = Stack.new(),
-			}
-			for id, node in pairs(match) do
-				local name = query.captures[id]
-				if name == 'container' then
-					match_record.container = node
-				else
-					if match_record[name] then match_record[name]:push(node) end
-				end
-			end
 
-			for _, other in matches:iter() do
-				if not ts.is_ancestor(match_record.container, other.container) then
-					break
+		for qid, node, _ in query:iter_captures(root_node, bufnr, start_row, end_row) do
+			local name = query.captures[qid]
+			-- check for 'delimiter' first, since that should be the most
+			-- common name
+			if name == 'delimiter' and match_record then
+				match_record.delimiter:push(node)
+			elseif name == 'container' and not match_record then
+				match_record = new_match_record()
+			elseif name == 'container' then
+				-- temporarily push the match_record to matches to be retrieved
+				-- later, since we haven't closed it yet
+				matches:push(match_record)
+				match_record = new_match_record()
+				-- since we didn't close the previous match_record, it must
+				-- mean that the current match_record has it as an ancestor
+				match_record.has_ancestor = true
+			elseif name == 'sentinel' and match_record then
+				-- if we see the sentinel, then we are done with the current
+				-- container
+				if match_record.has_ancestor then
+					local prev_match_record = matches:pop()
+					if prev_match_record then
+						-- since we have an ancestor, it has to be the last
+						-- element of the stack
+						prev_match_record.children:push(match_record)
+						match_record = prev_match_record
+					else
+						-- since match_record.has_ancestor was true, we shouldn't
+						-- be able to get to here unless something went wrong
+						-- with the queries or treesitter itself
+						log.error([[You are missing a @container,
+									which should be impossible!
+									Please double check the queries.]])
+					end
+				else
+					-- if match_record doesn't have an ancestor, the sentinel
+					-- means that we are done with it
+					matches:push(match_record)
+					match_record = nil
 				end
-				match_record.children:push(other)
-				matches:pop()
-			end
-			matches:push(match_record)
+			elseif (name == 'delimiter' or name == 'sentinel') and not match_record then
+				log.error([[You query got the capture name %s.
+					But it didn't come with a container, which should be impossible!
+						Please double check your queries.]], name)
+			end -- do nothing with other capture names
 		end
 	end
 
@@ -101,6 +143,7 @@ end
 
 ---Update highlights for every tree in given buffer.
 ---@param bufnr number # Buffer number
+---@param parser LanguageTree
 ---@return nil
 local function full_update(bufnr, parser)
 	log.debug('Performing full updated on buffer %d', bufnr)
@@ -112,10 +155,16 @@ local function full_update(bufnr, parser)
 	parser:for_each_tree(callback)
 end
 
+
 ---Sets up all the callbacks and performs an initial highlighting
-local function setup_parser(bufnr, parser)
+---@param bufnr number # Buffer number
+---@param parser LanguageTree
+---@param start_parent_lang string? # Parent language or nil
+---@return nil
+local function setup_parser(bufnr, parser, start_parent_lang)
 	log.debug('Setting up parser for buffer %d', bufnr)
-	util.for_each_child(parser:lang(), parser, function(p, lang)
+
+	util.for_each_child(start_parent_lang, parser:lang(), parser, function(p, lang, parent_lang)
 		log.debug("Setting up parser for '%s' in buffer %d", lang, bufnr)
 		-- Skip languages which are not supported, otherwise we get a
 		-- nil-reference error
@@ -123,36 +172,75 @@ local function setup_parser(bufnr, parser)
 
 		p:register_cbs {
 			on_changedtree = function(changes, tree)
-				if #changes == 0 then
-					-- Hack: an empty change set results from an undo.  Force a
-					-- full update by setting the maximum possible range
-					table.insert(changes, {tree:root():range()})
-				end
 				log.trace('Changed tree in buffer %d with languages %s', bufnr, lang)
 				-- HACK: As of Neovim v0.9.1 there is no way of unregistering a
 				-- callback, so we use this check to abort
 				if not lib.buffers[bufnr] then return end
 
-				-- Normalize the changes objects
-				changes = vim.tbl_map(normalize_change, changes)
+				-- HACK: changes can accidentally overwrite highlighting in injected code
+				-- blocks.
+				if not parent_lang then
+					-- If we have no parent language, then we use changes, otherwise we use the
+					-- whole tree's range.
+					-- Normalize the changes object if we have no parent language (the one we
+					-- get from on_changedtree)
+					changes = vim.tbl_map(normalize_change, changes)
+				elseif parent_lang ~= lang and changes[1] then
+					-- We have a parent language, so we are in an injected language code
+					-- block, thus we update all of the current code block
+					changes = {{tree:root():range()}}
+				else
+					-- some languages (like rust) use injections of the language itself for
+					-- certain functionality (e.g., macros in rust).  For these the
+					-- highlighting will be updated by the non-injected language part of the
+					-- code.
+					changes = {}
+				end
 
-				-- If a line has been moved from another region it will still
-				-- carry with it the extmarks from the old region.  We need to
-				-- clear all extmarks which do not belong to the current
-				-- language
+				-- If a line has been moved from another region it will still carry with it
+				-- the extmarks from the old region.  We need to clear all extmarks which
+				-- do not belong to the current language
 				for _, change in ipairs(changes) do
 					for key, nsid in pairs(lib.nsids) do
 						if key ~= lang then
-							vim.api.nvim_buf_clear_namespace(bufnr, nsid, change[1], change[3])
+							-- HACK: changes in the main language sometimes need to overwrite
+							-- highlighting on one more line
+							local line_end = change[3] + (parent_lang and 0 or 1)
+							vim.api.nvim_buf_clear_namespace(bufnr, nsid, change[1], line_end)
 						end
 					end
 				end
-				update_range(bufnr, changes, tree, lang)
+
+				-- only update highlighting if we have changes
+				if changes[1] then
+					update_range(bufnr, changes, tree, lang)
+				end
+
+				-- HACK: Since we update the whole tree when we have a parent
+				-- language, we need to make sure to then update all children
+				-- too, even if there is no change in them. This shouldn't
+				-- affect performance, since it only affects code nested at
+				-- least 2 injection languages deep.
+				if parent_lang then
+					local children = p:children()
+					for child_lang, child in pairs(children) do
+						if lang == child_lang then return end
+						child:for_each_tree(function(child_tree, child_p)
+							local child_changes = {{child_tree:root():range()}}
+
+							--  we don't need to remove old extmarks, since
+							--  the above code will handle that correctly
+							--  already, but we might have accidentally
+							--  removed extmarks that we need to set again
+							update_range(bufnr, child_changes, child_tree, child_p:lang())
+						end)
+					end
+				end
 			end,
 			-- New languages can be added into the text at some later time, e.g.
 			-- code snippets in Markdown
 			on_child_added = function(child)
-				setup_parser(bufnr, child)
+				setup_parser(bufnr, child, lang)
 			end,
 		}
 		log.trace("Done with setting up parser for '%s' in buffer %d", lang, bufnr)
@@ -165,10 +253,10 @@ end
 function M.on_attach(bufnr, settings)
 	log.trace('global strategy on_attach')
 	local parser = settings.parser
-	setup_parser(bufnr, parser)
+	setup_parser(bufnr, parser, nil)
 end
 
-function M.on_detach(bufnr)
+function M.on_detach(_bufnr)
 end
 
 function M.on_reset(bufnr, settings)
--- a/lua/rainbow-delimiters/strategy/local.lua
+++ b/lua/rainbow-delimiters/strategy/local.lua
@@ -35,20 +35,23 @@ local M = {}
 -- match tree every time the cursor moves.
 
 
--- Cache of match trees. We compute the match tree on every change, so that
--- when the cursor moves without changing the tree we don't need to re-compute
--- it.
+---Cache of match trees, maps a buffer number to its match tree.  We compute
+---the match tree on every change, so that when the cursor moves without
+---changing the tree we don't need to re-compute it.
+---
+---Each match tree maps a language and TS Tree to the corresponding match tree.
+---We need TS Tree because there might be multiple trees per buffer, such as a
+---Markdown buffer which contains multiple code blocks.
 local match_trees = {}
 
--- Reusable autogroup for events in this strategy.
+---Reusable autogroup for events in this strategy.
+---@type number
 local augroup = api.nvim_create_augroup('TSRainbowLocalCursor', {})
 
 
 ---Highlights a single match with the given highlight group
 local function highlight_match(bufnr, lang, match, hlgroup)
-	for _, opening      in match.opening:iter()      do lib.highlight(bufnr, lang, opening,      hlgroup) end
-	for _, closing      in match.closing:iter()      do lib.highlight(bufnr, lang, closing,      hlgroup) end
-	for _, intermediate in match.intermediate:iter() do lib.highlight(bufnr, lang, intermediate, hlgroup) end
+	for _, delimiter in match.delimiter:iter() do lib.highlight(bufnr, lang, delimiter, hlgroup) end
 end
 
 ---Highlights all matches and their children on the stack of matches. All
@@ -75,43 +78,89 @@ local function find_container(matches, container, level)
 	end
 end
 
+--- Create a new empty match_record with an optionally set container
+---@param container TSNode?
+---@return table
+local function new_match_record(container)
+	return {
+		container = container,
+		delimiter = Stack.new(),
+		children = Stack.new(),
+	}
+end
+
 ---Assembles the match tree, usually called after the document tree has
 ---changed.
+---@param bufnr   number  Buffer number
+---@param changes table   List of node ranges in which the changes occurred
+---@param tree    TSTree  TS tree
+---@param lang    string  Language
+---@return Stack?
 local function build_match_tree(bufnr, changes, tree, lang)
+	if not lib.enabled_for(lang) then return end
+
 	local query = lib.get_query(lang)
+	if not query then return end
+
 	local matches = Stack.new()
 
 	for _, change in ipairs(changes) do
+		-- This is the match record, it lists all the relevant nodes from
+		-- each match.
+		---@type table?
+		local match_record
 		local root_node = tree:root()
-		for _, match, _ in query:iter_matches(root_node, bufnr, change[1], change[3] + 1) do
-			-- This is the match record, it lists all the relevant nodes from
-			-- the match.
-			local match_record = {
-				opening = Stack.new(),
-				closing = Stack.new(),
-				intermediate = Stack.new(),
-				children = Stack.new(),
-			}
-			for id, node in pairs(match) do
-				local name = query.captures[id]
-				if name == 'container' then
-					match_record.container = node
+		local start_row, end_row = change[1], change[3] + 1
+		lib.clear_namespace(bufnr, lang, start_row, end_row)
+		for qid, node, _ in query:iter_captures(root_node, bufnr, start_row, end_row) do
+			local name = query.captures[qid]
+			-- check for 'delimiter' first, since that should be the most
+			-- common name
+			if name == 'delimiter' and match_record then
+				match_record.delimiter:push(node)
+			elseif name == 'container' and not match_record then
+				match_record = new_match_record(node)
+			elseif name == 'container' then
+				local prev_match_record = match_record
+				-- temporarily push the match_record to matches to be retrieved
+				-- later, since we haven't closed it yet
+				matches:push(match_record)
+				match_record = new_match_record(node)
+				-- since we didn't close the previous match_record, it must
+				-- mean that the current match_record has it as an ancestor
+				match_record.ancestor = prev_match_record
+			elseif name == 'sentinel' and match_record then
+				-- if we see the sentinel, then we are done with the current
+				-- container
+				if match_record.ancestor then
+					local prev_match_record = matches:pop()
+					if prev_match_record then
+						-- since we have an ancestor, it has to be the last
+						-- element of the stack
+						prev_match_record.children:push(match_record)
+						match_record = prev_match_record
+					else
+						-- since match_record.has_ancestor was true, we shouldn't
+						-- be able to get to here unless something went wrong
+						-- with the queries or treesitter itself
+						log.error([[You are missing a @container,
+									which should be impossible!
+									Please double check the queries.]])
+					end
 				else
-					if match_record[name] then match_record[name]:push(node) end
+					-- if match_record doesn't have an ancestor, the sentinel
+					-- means that we are done with it
+					matches:push(match_record)
+					match_record = nil
 				end
-			end
-
-			for _, other in matches:iter() do
-				if not ts.is_ancestor(match_record.container, other.container) then
-					break
-				end
-				other.ancestor = match_record
-				match_record.children:push(other)
-				matches:pop()
-			end
-			matches:push(match_record)
+			elseif (name == 'delimiter' or name == 'sentinel') and not match_record then
+				log.error([[You query got the capture name: %s.
+					But it didn't come with a container, which should be impossible!
+					Please double check your queries.]], name)
+			end -- do nothing with other capture names
 		end
 	end
+
 	return matches
 end
 
@@ -126,7 +175,7 @@ local function update_local(bufnr, tree, lang)
 		local curpos = api.nvim_win_get_cursor(0)
 		-- The order of traversal guarantees that the first match which
 		-- contains the cursor is also the lowest one.
-		for _, match in query:iter_matches(tree:root(), bufnr) do
+		for _, match in query:iter_matches(tree:root(), bufnr, 0, -1) do
 			if cursor_container then break end
 			for id, node in pairs(match) do
 				local name = query.captures[id]
@@ -139,8 +188,16 @@ local function update_local(bufnr, tree, lang)
 	end
 	if not cursor_container then return end
 
-	local matches = match_trees[bufnr][lang]
-	if not matches then return end
+	local matches_lang = match_trees[bufnr][lang]
+	if not matches_lang then
+		log.debug("Did not build any matches Stack for language '%s'", lang)
+		return
+	end
+	local matches = matches_lang[tree]
+	if not matches then
+		log.debug("Did not build any matches Stack for tree '%s'", vim.inspect(tree:range()))
+		return
+	end
 
 	-- Find the correct container in the match tree
 	local cursor_match, level = find_container(matches, cursor_container, 1)
@@ -151,7 +208,8 @@ local function update_local(bufnr, tree, lang)
 
 	-- Starting with the cursor match travel up and highlight every ancestor as
 	-- well
-	local ancestor, level = cursor_match.ancestor, level - 1
+	local ancestor = cursor_match.ancestor
+	level = level - 1
 	while ancestor do
 		highlight_match(bufnr, lang, ancestor, lib.hlgroup_at(level))
 		ancestor, level = ancestor.ancestor, level - 1
@@ -167,9 +225,12 @@ local function local_rainbow(bufnr, parser)
 end
 
 ---Sets up all the callbacks and performs an initial highlighting
+---@param bufnr number # Buffer number
+---@param parser LanguageTree
+---@return nil
 local function setup_parser(bufnr, parser)
 	log.debug('Setting up parser for buffer %d', bufnr)
-	util.for_each_child(parser:lang(), parser, function(p, lang)
+	util.for_each_child(nil, parser:lang(), parser, function(p, lang, _parent_lang)
 		log.debug("Setting up parser for '%s' in buffer %d", lang, bufnr)
 		-- Skip languages which are not supported, otherwise we get a
 		-- nil-reference error
@@ -187,9 +248,10 @@ local function setup_parser(bufnr, parser)
 				local fake_changes = {
 					{tree:root():range()}
 				}
-				match_trees[bufnr][lang] = build_match_tree(bufnr, fake_changes, tree, lang)
+				match_trees[bufnr][lang] = match_trees[bufnr][lang] or {}
+				match_trees[bufnr][lang][tree] = build_match_tree(bufnr, fake_changes, tree, lang)
 				-- Re-highlight after the change
-				local_rainbow(bufnr, parser)
+				local_rainbow(bufnr, p)
 			end,
 			-- New languages can be added into the text at some later time, e.g.
 			-- code snippets in Markdown
@@ -209,7 +271,7 @@ function M.on_attach(bufnr, settings)
 		group = augroup,
 		buffer = bufnr,
 		callback = function(args)
-			util.for_each_child(parser:lang(), parser, function(_, lang)
+			util.for_each_child(nil, parser:lang(), parser, function(_, lang, _)
 				lib.clear_namespace(bufnr, lang)
 			end)
 			local_rainbow(args.buf, parser)
@@ -223,7 +285,8 @@ function M.on_attach(bufnr, settings)
 		local changes = {
 			{tree:root():range()}
 		}
-		match_trees[bufnr][sub_lang] = build_match_tree(bufnr, changes, tree, sub_lang)
+		match_trees[bufnr][sub_lang] = {}
+		match_trees[bufnr][sub_lang][tree] = build_match_tree(bufnr, changes, tree, sub_lang)
 	end)
 	local_rainbow(bufnr, parser)
 end
--- a/queries/elixir/rainbow-delimiters.scm
+++ b/queries/elixir/rainbow-delimiters.scm
@@ -1,31 +1,30 @@
 (call
   (arguments
-    "(" @opening
-    ")" @closing) @container)
+    "(" @delimiter
+    ")" @delimiter @sentinel) @container)
 
 (block
-  "(" @opening
-  ")" @closing) @container
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
 
-;; This pattern breaks the parser.
-; Error attaching strategy to buffer 1: /usr/share/nvim/runtime/lua/vim/treesitter/query.lua:259: query: invalid structure at position 170 for language elixir
 (string
   (interpolation
-    "#{" @opening
-    "}" @closing) @container)
+    "#{" @delimiter
+    "}" @delimiter @sentinel) @container)
 
 (tuple
-  "{" @opening
-  "}" @closing) @container
+  "{" @delimiter
+  "}" @delimiter @sentinel) @container
 
 (list
-  "[" @opening
-  "]" @closing) @container
+  "[" @delimiter
+  "]" @delimiter @sentinel) @container
 
 (bitstring
-  "<<" @opening
-  ">>" @closing) @container
+  "<<" @delimiter
+  ">>" @delimiter @sentinel) @container
 
 (map
-  "{" @opening
-  "}" @closing) @container
+  "%" @delimiter
+  "{" @delimiter
+  "}" @delimiter @sentinel) @container
--- a/queries/fennel/rainbow-delimiters.scm
+++ b/queries/fennel/rainbow-delimiters.scm
@@ -1,165 +1,124 @@
 (list
-  "(" @opening
-  ")" @closing) @container
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
 
 (table
-  "{" @opening
-  (":" @intermediate)*
-  "}" @closing) @container
+  "{" @delimiter
+  (":" @delimiter _)*
+  "}" @delimiter @sentinel) @container
 
 (sequential_table
-  "[" @opening
-  "]" @closing) @container
+  "[" @delimiter
+  "]" @delimiter @sentinel) @container
 
 (fn
-  "(" @opening
-  ")" @closing) @container
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
 
 (lambda
-  "(" @opening
-  ")" @closing) @container
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
 
 (let
-  "(" @opening
-  ")" @closing) @container
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
 
 (set
-  "(" @opening
-  ")" @closing) @container
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
 
 
-;;; BUG: regarding the next 6 pairs of queries:
-;;;      since both pairs of delimiters
-;;;      are direct children of the same `TSNode`
-;;;      they get highligthed in different colours
-
-;;; NOTE: https://github.com/neovim/neovim/pull/17099
+;;; NOTE: The following queries contain one set of outer parentheses and one
+;;;       set of inner parentheses, but both are on the same level of the tree.
+;;;       Thus we cannot match the inner ones one level deeper than the outer
+;;;       ones.
 
 (each
-  "(" @opening
-  ")" @closing) @container
-
-(each
-  "[" @opening
-  "]" @closing) @container
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
 
 (collect
-  "(" @opening
-  ")" @closing) @container
-
-(collect
-  "[" @opening
-  "]" @closing) @container
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
 
 (icollect
-  "(" @opening
-  ")" @closing) @container
-
-(icollect
-  "[" @opening
-  "]" @closing) @container
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
 
 ;;; TODO: add these when treesitter parser
 ;;;       adds support for fennel 1.1.1
 
 ;; (fcollect
-;;   "(" @opening
-;;   ")" @closing) @container
+;;   "(" @delimiter
+;;   ")" @delimiter @sentinel) @container
 ;;
 ;; (fcollect
-;;   "[" @opening
-;;   "]" @closing) @container
+;;   "[" @delimiter
+;;   "]" @delimiter @sentinel) @container
 
 (accumulate
-  "(" @opening
-  ")" @closing) @container
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
 
 (accumulate
-  "[" @opening
-  "]" @closing) @container
+  "[" @delimiter
+  "]" @delimiter @sentinel) @container
 
 ;;; TODO: add these when treesitter parser
 ;;;       adds support for fennel 1.3.0
 
 ;; (faccumulate
-;;   "(" @opening
-;;   ")" @closing) @container
+;;   "(" @delimiter
+;;   ")" @delimiter @sentinel) @container
 ;;
 ;; (faccumulate
-;;   "[" @opening
-;;   "]" @closing) @container
-
-;;; TODO: (reo101) swap out the above queries
-;;;       for these (below) when 17099 (see NOTE) is merged
-
-;; (each
-;;    ["[" ")"]+ @opening
-;;    ["]" ")"]+ @closing) @container
-;;
-;; (collect
-;;    ["[" ")"]+ @opening
-;;    ["]" ")"]+ @closing) @container
-;;
-;; (icollect
-;;    ["[" ")"]+ @opening
-;;    ["]" ")"]+ @closing) @container
-;;
-;; (fcollect
-;;    ["[" ")"]+ @opening
-;;    ["]" ")"]+ @closing) @container
-;;
-;; (accumulate
-;;    ["[" ")"]+ @opening
-;;    ["]" ")"]+ @closing) @container
-;;
-;; (faccumulate
-;;    ["[" ")"]+ @opening
-;;    ["]" ")"]+ @closing) @container
+;;   "[" @delimiter
+;;   "]" @delimiter @sentinel) @container
 
 (for
-  "(" @opening
-  ")" @closing) @container
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
 
 (for_clause
-  "[" @opening
-  "]" @closing) @container
+  "[" @delimiter
+  "]" @delimiter @sentinel) @container
 
 (var
-  "(" @opening
-  ")" @closing) @container
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
 
 (parameters
-  "[" @opening
-  "]" @closing) @container
+  "[" @delimiter
+  "]" @delimiter @sentinel) @container
 
 (let_clause
-  "[" @opening
-  "]" @closing) @container
+  "[" @delimiter
+  "]" @delimiter @sentinel) @container
 
 (quoted_list
-  "(" @opening
-  ")" @closing) @container
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
 
 (local
-  "(" @opening
-  ")" @closing) @container
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
 
 (multi_value_binding
-  "(" @opening
-  ")" @closing) @container
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
 
 (hashfn
-  "(" @opening
-  ")" @closing) @container
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
 
 (match
-  "(" @opening
-  ")" @closing) @container
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
 
 (sequential_table_pattern
-  "[" @opening
-  "]" @closing) @container
+  "[" @delimiter
+  "]" @delimiter @sentinel) @container
 
 (sequential_table_binding
-  "[" @opening
-  "]" @closing) @container
+  "[" @delimiter
+  "]" @delimiter @sentinel) @container
--- a/queries/html/rainbow-delimiters.scm
+++ b/queries/html/rainbow-delimiters.scm
@@ -1,19 +1,61 @@
-;;; A pair of opening and closing tag with any content in-between. Excludes
-;;; self-closing tags or opening tags without closing tag.
+;;; A pair of delimiter tags with any content in-between.
+;;; Last tag should be a sentinel.
 
 (element
-  (start_tag (tag_name) @opening)
-  ; (element (self_closing_tag) @intermediate)*
-  (end_tag (tag_name) @closing)) @container
+  (start_tag
+    "<" @delimiter
+    (tag_name) @delimiter
+    ">" @delimiter)
+  (end_tag
+    "</" @delimiter
+    (tag_name) @delimiter
+    ">" @delimiter @sentinel)) @container
 
 (element
-  (self_closing_tag (tag_name) @opening)) @container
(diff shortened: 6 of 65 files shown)


HiPhish:
Thank you for wanting to help out. Can you please explain how you worked around the issue of matching the same capture multiple times? My understanding is that if you have a query like `(foo (bar) @xxx (baz) @xxx)` Neovim's Tree-sitter API cannot handle the two `@xxx` captures and will only give me one. If I recall correctly that is why I have separate `@opening`, `@intermediate` and `@closing` captures.

Danielkonge:
The problem is not really Neovim from what I can tell. If you play around with the treesitter playground (https://github.com/nvim-treesitter/playground) all the queries captures multiple captures fine. 

Screenshot 2023-10-07 at 18 40 57

(Look at the query and lines 13 to 21.)

The problem is really with `iter_matches` (which is the obvious choice for this algorithm). (This is the problem discussed here: https://github.com/neovim/neovim/pull/17099 )

So I tried rewriting basically the same algorithm but using `iter_captures` instead of `iter_matches`, since `iter_captures` does have the behavior I would expect (unlike `iter_matches`). 

Unlike `iter_matches`, which captures the `@container` as a whole and then only can give you one of each capture inside the `@container` (e.g. only one `@intermediate` above), `iter_captures` gets all the captures in the following order:

`@container` first and then the order the other captures show up in in the code. If there is another `@container` inside the first `@container`, then we start going through that. I.e., we go through all captures in the order they show up in the code, but each "large" capture starts with `@container` whereafter we go through the captures inside the `@container`. 

The algorithm tries to mirror the previous algorithm (depth first search). 

(1) We make an empty `matches` stack and an empty `match_record` and start going through all the captured nodes based on our queries (on the changed part as before).

(2) If the captures name is `@container` that means we are starting going through a new container. 

- If `match_record.container` is not yet set, then that means that we are not currently inside another `@container`, so we set `match_record.container = node`. 

- If `match_record.container` is set, then that means that our current node is inside `match_record.container` and we starting on a new container. So we (temporarily) push `match_record` to `matches` so we can pop it back out later when we are done going through the current node. And then we start a new `match_record` with `container` equal to the current node.

- If the node's capture name is `@closing` that means that we are done with the current container (this is a new assumption I added!). So we set `match_record.closing` to be the current node and pop out the previous match_record from `matches` (if there is one).

    * If the previous match_record exists and is an ancestor of the current `match_record`, then we add the current `match_record` to the `children` of the previous match_record, which we are still inside (the one we temporarily pushed earlier). And we continue working with the previous match_record which is now the new current `match_record` since we are still working inside it.

    * Otherwise, if the previous match_record exists but is not an ancestor of the current `match_record`, then we are done with the current `match_record` (since we saw its `@closing`), so after popping out the previous match_record, we push both the current `match_record` and the previous match_record to `matches` (since it was not an ancestor, we are out of it too). [The order we push these doesn't matter.] Finally we create a new empty `match_record` in this case.

  - If there is no previous match_record, we cannot be inside another `@container`, so we just push the current `match_record` to `matches` and create a new empty `match_record`.

- If the node's capture name is not `@container` or `@closing`, then we add to the corresponding field in `match_record` (`@opening` or `@intermediate`) if such a field exists. Since we haven't seen a `@closing` in the current `@container` yet, this will be assigned correctly simply by the order `iter_captures` goes through the captures. 



All of this is really just trying to implement the old algorithm with the restrictions we get from using `iter_caputes` instead of `iter_matches`. 

If we want this to replace the current `global` (and maybe `local`?) strategy, I think it would be a good idea to actually still keep the old code around and just comment it out (this pull request doesn't do this right now). If they ever fix `iter_matches` so it has the expected behavior upstream, I think the current implementation is nicer than what you see in this pull request. [It doesn't add the extra assumption that all `@container`s end with a single `@closing`.]

Another solution could be to keep a better version of `iter_matches` as part of rainbow-delimiters if you prefer that instead.

Danielkonge:
Screenshot 2023-10-07 at 20 31 20


This might be an even better example showing that Neovim can catch all the `test` in the pattern although they all have the same name here. 

(Line 13-21 again.)

HiPhish:
Ah yes, the cursed issue where everyone who takes it over vanishes. I am familiar with `iter_matches`, but I did not consider using `iter_captures`.

I will have to go over your algorithm in detail when I have the time one of these days. But I don't like that `@closing` has to be a single node, this just kicks the can further down the road. How about this:

- there are only two captures `@container` and `@delimiter`
- As you have stated a new `@capture` starts a new context
- For every `@delimiter` node check whether it is a child of the current topmost `@container` on the stack
  - If it is then keep going
  - Otherwise we are done with the current `@container`

This avoids the need for a `@closing` node altogether, but in return we have to perform this check for every single node. I don't know if this would make much of a performance impact or if it is just a matter of comparing a few numbers. What do you think of my proposal?

> If we want this to replace the current `global` (and maybe `local`?) strategy, I think it would be a good idea to actually still keep the old code around and just comment it out (this pull request doesn't do this right now).

They would not work if we rename the captures in the queries. We could keep the old names and just treat them the same in the new strategies. But honestly I don't see much of a point in keeping the old strategies around. The algorithm is limited to one function and described in the HACKING file if we need it for posterity.

On a side note, how hard did you find it to get into the code? I rarely get non-trivial contributions, so I have no idea how readable and approachable my code really is.

Danielkonge:
First a note to the algorithm I wrote above: You can actually avoid doing the check for whether something is an ancestor (where I wrote **NOTE** above) if you assume that there is only one `@closing`, since you can just check instead if the previous match record has already been closed or not. I don't know if this will actually speed up things significantly or not, but I thought I would note it for now.


> I will have to go over your algorithm in detail when I have the time one of these days. But I don't like that `@closing` has to be a single node, this just kicks the can further down the road. How about this:
> 
> * there are only two captures `@container` and `@delimiter`
> * As you have stated a new `@capture` starts a new context
> * For every `@delimiter` node check whether it is a child of the current topmost `@container` on the stack
>   
>   * If it is then keep going
>   * Otherwise we are done with the current `@container`
> 
> This avoids the need for a `@closing` node altogether, but in return we have to perform this check for every single node. I don't know if this would make much of a performance impact or if it is just a matter of comparing a few numbers. What do you think of my proposal?

Your suggestion is definitely possible and it shouldn't be that hard to change what I currently have. I can try to have code for both algorithms written so we can test them a bit more (this sounds like something where we might want to actually benchmark the performance of these things, since I also don't really know the implications of checking if something is a parent of a treesitter node or not). With your suggestion we will probably still need the check whether the previous match record is an ancestor of the current match record though. [I think the parent check should be quick, but the ancestor check is probably not as quick.]
 
> > If we want this to replace the current `global` (and maybe `local`?) strategy, I think it would be a good idea to actually still keep the old code around and just comment it out (this pull request doesn't do this right now).
> 
> They would not work if we rename the captures in the queries. We could keep the old names and just treat them the same in the new strategies. But honestly I don't see much of a point in keeping the old strategies around. The algorithm is limited to one function and described in the HACKING file if we need it for posterity.

I will start out by keeping the old capture names around for now, just because we would need to edit all the queries otherwise, and it will also make it easier to compare the two implementations. If we decide on changing the names to `@delimiter` this will be easy to change later. 
 
> On a side note, how hard did you find it to get into the code? I rarely get non-trivial contributions, so I have no idea how readable and approachable my code really is.

I think it was pretty readable, but I mostly looked at the queries and strategies (+ a little at the general implementation). I am pretty new to Lua, so some of the meta tables stuff was a bit confusing, but that probably has more to do with me not being used to Lua meta tables yet than your actual code. Also I have never used Vader, so for testing I have just been manually checking whether it works as expected in my day to day programming + a few of the test files available. 

To be a bit more precise, I think your Stack implementation and the strategy code is very easy to follow, but I was a bit confused about how the namespaces worked in `lib.lua`. After looking at it more, I guess part of it is because `vim.api.nvim_create_namespace()` doesn't behave as I would have liked. (I would have liked it to be easier to have per-language namespaces that I could search for when looking at Extmarks.)

Danielkonge:
I have added an alternative `update_range` function (currently called `update_range2`) that is easy to change so it works with just `@delimiter` and `@container` (all that is needed is changing the names of capture groups in `new_match_record` and `highlight_matches`). I have only tested functionality and not performance so far, but it seems to work fine and I can't tell any difference in performance on my computer. Though we should probably do some actual performance tests comparing `update_range1` with `update_range2`.

Danielkonge:
I have a large local Rust file (5.5MB) that I tried to use to compare the two algorithms. 

On my computer with this file, both algorithms seem to highlight ok, and the main part of `update_range1` takes about 0.6secs on average while the main part of `update_range2` takes about 1.2secs. This doesn't seem like a huge difference, but `update_range1` was consistently about twice as fast as `update_range2`. 

Also, it should be noted that this file has almost no nested highlights levels (it is basically all level 2 in one long list), so `ts.is_ancestor` only needs to do very little work. 

In general, it seems that keeping track of `@closing` tags and ensuring that there is exactly one `@closing` for every `@container` will be faster, but for more common size files there is no difference. (On smaller normal size files they both take about 0.01-0.05secs for me.)

Also note that this was timed with `os.clock()` in Lua, which I can read is not the best for real world timing of algorithms.

HiPhish:
There is also a C++ file under `test/stress/markdown/lorem-ipsum.md` which you can try. It has a number of languages embedded in Markdown, but it might be too short to deliver any meaningful results.

If we want to go with the `@closing` capture then how about we treat it purely as a sentinel node? By that I mean the content which we want to highlight is captured with `@delimiter`, but in addition to that we also capture the last delimiter with `@closing`.  Here is what it would look like for Lua function definitions where the sentinel is also the last delimiter:

```
(function_declaration
  "function" @delimiter
  "end" @delimiter @closing) @container
```

Here is a contrived HTML example in which for whatever reason I only want to highlight the name of the closing tag, so the delimiter is `tag_name` and the sentinel is `end_tag`:

```
(element
  (start_tag
    "<" @delimiter
    (tag_name) @delimiter
    ">" @delimiter)
  (end_tag
    (tag_name) @delimiter) @closing) @container
```

This will require adjusting queries, but the change is straight-forward. I can do that myself, no need to waste your time.

Danielkonge:
> There is also a C++ file under `test/stress/markdown/lorem-ipsum.md` which you can try. It has a number of languages embedded in Markdown, but it might be too short to deliver any meaningful results.

Both algorithms take 0.00secs in my timing, so it is definitely too short to see anything.  (Though the timing is not that accurate since right now it times separately for each language in the file.)

> If we want to go with the `@closing` capture then how about we treat it purely as a sentinel node? By that I mean the content which we want to highlight is captured with `@delimiter`, but in addition to that we also capture the last delimiter with `@closing`. Here is what it would look like for Lua function definitions where the sentinel is also the last delimiter:
> 
> ```
> (function_declaration
>   "function" @delimiter
>   "end" @delimiter @closing) @container
> ```

That would definitely be a possibility too. 

> Here is a contrived HTML example in which for whatever reason I only want to highlight the name of the closing tag, so the delimiter is `tag_name` and the sentinel is `end_tag`:
> 
> ```
> (element
>   (start_tag
>     "<" @delimiter
>     (tag_name) @delimiter
>     ">" @delimiter)
>   (end_tag
>     (tag_name) @delimiter) @closing) @container
> ```
> 
> This will require adjusting queries, but the change is straight-forward. I can do that myself, no need to waste your time.

I can easily edit the code to work like that very quickly if this is what we want to go with. I think this is a good idea since the code with `@closing` is also simpler (so hopefully easier to maintain), and it makes sense that the capture starts with `@container` and ends with `@closing`. And based on my (admittedly very simple) testing it does seem to be faster than calling `ts.is_ancestor()`.

HiPhish:
BTW, am not particularly attached to the capture names, so if you have a better suggestions don't hesitate. For example, maybe `@sentinel` would be better than `@closing`, because the name `@closing` makes me think there needs to be a corresponding `@opening` somewhere.

Danielkonge:
I think `@sentinel` does better describe the idea, but it is a pretty uncommon word. Then again, it might not be bad to use unique and descriptive words, so for now let's go with `@sentinel`.

Danielkonge:
I have updated the algorithm so it should work with `@delimiter` and `@sentinel` as described now. I have checked that it works as I would expect with Lua and Rust so far (and I even managed to speed it up a little [maybe ~10%] by pushing and popping less).

Danielkonge:
I have tried to update the local strategy in a similar way now too. I think it works as expected, but I only really use the global strategy normally, so I am not entirely sure if it used to behave slightly different in some cases.

Danielkonge:
I have been testing injected languages in `test/stress/markdown/lorem-ipsum.md`, but the following lines in the original code seems to cause problems:

https://github.com/HiPhish/rainbow-delimiters.nvim/blob/652345bd1aa333f60c9cbb1259f77155786e5514/lua/rainbow-delimiters/strategy/global.lua#L139C1-L149C8

For now I have commented out those lines and don't see anything going wrong, but I assume that was put there for a reason. @HiPhish do you remember an example where this code is clearly needed?

Also note: I updated the queries for the corresponding languages I needed to test with that markdown file. It seems to work as expected after commenting out the above lines.

HiPhish:
Yes, I remember why this exists. Let's say you have a Markdown file with embedded Lua:

```markdown
Here is some Lua code:

~~~lua
-- This is Lua
print({{{}}})
~~~

Back to Markdown
```

Now if we *move* (not delete and put) the Lua code one line down it will carry with it the extmarks it had and we will have Lua delimiter highlighting within Markdown. This is because extmarks move with the text. My solution is to have per-language namespaces and delete all extmarks which do not belong to the current language. In the above example the line has Lua extmarks, so when I move it to the Markdown region all non-Markdown extmarks are removed.

Danielkonge:
> Yes, I remember why this exists. Let's say you have a Markdown file with embedded Lua:
> 
> ```
> Here is some Lua code:
> 
> ~~~lua
> -- This is Lua
> print({{{}}})
> ~~~
> 
> Back to Markdown
> ```
> 
> Now if we _move_ (not delete and put) the Lua code one line down it will carry with it the extmarks it had and we will have Lua delimiter highlighting within Markdown. This is because extmarks move with the text. My solution is to have per-language namespaces and delete all extmarks which do not belong to the current language. In the above example the line has Lua extmarks, so when I move it to the Markdown region all non-Markdown extmarks are removed.

I see. I am able to reproduce this sometimes (in some languages). But the problems I have while enabling that code is worse. 

Without the code, I can get the problem like this:

https://github.com/HiPhish/rainbow-delimiters.nvim/assets/7075380/d7e646cc-55dc-44e6-8015-4d895d1da93f

With the code, I can't update the highlighting as usual (and sometimes the highlighting starts out wrong). If I try to do `:e<CR>`, then highlighting will go away while visible, and I need to scroll away and do `:e<CR>` and then scroll back to get it to work. 

https://github.com/HiPhish/rainbow-delimiters.nvim/assets/7075380/95a7c4a5-0918-48d2-9b31-35fd8ed8bb4e

I will try to see if I can find a better way to do this that solves both issues.

HiPhish:
Yes, I have noticed that the `changes` can be quite weird at time. The number `4294967295` is the maximum possible 32-bit unsigned integer value. Neovim uses the range `(0, 4294967295)` to mean "the entire buffer".

Can you please mark your PR as WIP or something like that until it is ready for review? I really appreciate your effort and I would prefer not to waste either of our time by reviewing code that isn't done yet.

Danielkonge:
> Yes, I have noticed that the `changes` can be quite weird at time. The number `4294967295` is the maximum possible 32-bit unsigned integer value. Neovim uses the range `(0, 4294967295)` to mean "the entire buffer".

Other than that number, I think the weirder thing is that it sometimes sends the same language range multiple times. E.g. in my list above, both markdown lists cover the entire buffer and the vim lists are identical. 

> Can you please mark your PR as WIP or something like that until it is ready for review? I really appreciate your effort and I would prefer not to waste either of our time by reviewing code that isn't done yet.

I have changed the PR to "Draft" in GitHub for now while I work on this last thing. I don't plan to make more changes to `update_range` in the global strategy though, so if you want to start looking at the main part of this pull request, that part should be final now.

Danielkonge:
@HiPhish: I managed to fix all issues I had with moving code, using undo and any other problems I saw now. If you notice any other problems I can try to fix those too, but otherwise I think this code is ready for review now. (I don't plan to change anything unless someone notices a problem.)

HiPhish:
Just a quick heads-up, I am in the process of reviewing this PR. Can you please tell me in which order captures are delivered by iter_capture?

HiPhish (changes requested):
I have performed a preliminary code review of some surface level stuff. I still need to understand the algorithm inside `update_range` of the global strategy. From what I gather there is the stack (`matches`) the current match record (`match_record`) that is being built up. We iterate through all the captures and differentiate based on the name of the capture:

- `delimiter` is obvious
- `container` has two cases:
  - If the current record has no container node, then this is its node
  - If the current match has been closed it belongs to a sibling match. Push the current match to the stack and start a new current match
  - Otherwise the container belongs to a new child match. Stash the current record on the stack and start a new record which has the previous one as its parent
- `sentinel` closes the current record
  - If the current record has no parent push it to the stack and start a new record
  - Otherwise pop the parent off the stack and set it as the current record

The `matches` stack serves two purposes: eventually it will be a "list" of top-level records, but until then it serves as a stash for records while we collect their children. Is my understanding so far correct?

Regarding the order of matches, my understanding is that all captures are returned in a depth-first order. So given this HTML snippet:

```html
<div id="d1">
	<div id="d2">
		<div id="d3">
		</div>
	</div>	
	<div id="d4">
	</div>	
</div>
```

We would get the captures in this order:

- container `d1`
- start `d1`
- container `d2`
- start `d2`
- container  `d3`
- start `d3`
- end `d3`
- sentinel `d3`
- end tag `d2`
- sentinel `d2`
- container `d4`
- start `d4`
- end `d4`
- sentinel`d4`
- end `d1`
- sentinel `d1`

Is this the correct order?

lua/rainbow-delimiters/strategy/global.lua
```diff
@@ -64,36 +78,63 @@ local function update_range(bufnr, changes, tree, lang)
...
+						-- with the queries or treesitter itself
+						log.debug([[You are missing a @container,
+									which should be impossible!
+									Please double check the queries.]])
+					end
+				else
+					-- if match_record doesn't have an ancestor, the sentinel
+					-- means that we are done with it
+					matches:push(match_record)
+					match_record = new_match_record()
```
Danielkonge, line 131: This can be changed to
```suggestion
					match_record = nil
```
if we change the above code as commented. This might be make the code easier to follow?

lua/rainbow-delimiters/strategy/global.lua
```diff
@@ -64,36 +78,63 @@ local function update_range(bufnr, changes, tree, lang)
...
-				else
-					if match_record[name] then match_record[name]:push(node) end
+		for qid, node, _ in query:iter_captures(root_node, bufnr, start_row, end_row) do
+			local name = query.captures[qid]
+			-- check for 'delimiter' first, since that should be the most
+			-- common name
+			if name == 'delimiter' then
+				match_record.delimiter:push(node)
+			elseif name == 'container' and not match_record.container then
+				match_record.container = node
```
Danielkonge, line 94: With the below change this can be changed to
```suggestion
			elseif name == 'container' and not match_record then
				match_record = new_match_record(node)
```

Danielkonge:
> I have performed a preliminary code review of some surface level stuff. I still need to understand the algorithm inside `update_range` of the global strategy. From what I gather there is the stack (`matches`) the current match record (`match_record`) that is being built up. We iterate through all the captures and differentiate based on the name of the capture:
> 
> * `delimiter` is obvious
> * `container` has two cases:
>   
>   * If the current record has no container node, then this is its node
>   * If the current match has been closed it belongs to a sibling match. Push the current match to the stack and start a new current match
>   * Otherwise the container belongs to a new child match. Stash the current record on the stack and start a new record which has the previous one as its parent
> * `sentinel` closes the current record
>   
>   * If the current record has no parent push it to the stack and start a new record
>   * Otherwise pop the parent off the stack and set it as the current record
> 
> The `matches` stack serves two purposes: eventually it will be a "list" of top-level records, but until then it serves as a stash for records while we collect their children. Is my understanding so far correct?

Yes, it sounds like you have the right idea. 

Looking back at the code, we can change the check for whether there is a current `container` to whether there is a current `match_record`. This complicates the stylua types (since we need to consider nil `match_record`s then), but it might be more clear?

> Regarding the order of matches, my understanding is that all captures are returned in a depth-first order. So given this HTML snippet:
> 
> ```
> <div id="d1">
> 	<div id="d2">
> 		<div id="d3">
> 		</div>
> 	</div>	
> 	<div id="d4">
> 	</div>	
> </div>
> ```
> 
> We would get the captures in this order:
> 
> * container `d1`
> * start `d1`
> * container `d2`
> * start `d2`
> * container  `d3`
> * start `d3`
> * end `d3`
> * sentinel `d3`
> * end tag `d2`
> * sentinel `d2`
> * container `d4`
> * start `d4`
> * end `d4`
> * sentinel`d4`
> * end `d1`
> * sentinel `d1`
> 
> Is this the correct order?

Yes, that is the correct order from my understanding of `iter_captures()`.

HiPhish:
I like the variant with `match_record = nil` better. Then we don't even need the `container` field in the match record table. I feel there is a really elegant recursive mathematical definition hidden somewhere under that imperative code, but let's get this out of the door as is first or else it will never get done :)

Can you please check the "allow edits from maintainers" checkbox in your PR? I have a few improvements to discuss that I want to push to this PR as a commit. If I can commit it will be more efficient than if I can only tell you what to do.

On a related note, I have played around with the HTML query and I absolutely love the result. Finally I can have the tag name and the angle brackets highlighted, but not the attributes. I have literally wanted this for years, ever since I first learned about rainbow parentheses. Thank you so much for your work!

Screenshot_20231025_213157

Danielkonge:
> I like the variant with `match_record = nil` better. Then we don't even need the `container` field in the match record table. I feel there is a really elegant recursive mathematical definition hidden somewhere under that imperative code, but let's get this out of the door as is first or else it will never get done :)

I have updated the code so it uses `match_record = nil` and removed `container` now.

> Can you please check the "allow edits from maintainers" checkbox in your PR? I have a few improvements to discuss that I want to push to this PR as a commit. If I can commit it will be more efficient than if I can only tell you what to do.

From my side it looks like this is already checked. Can you not edit now? I will try to see if I need to click "Yes" to something in some settings somewhere. 

> On a related note, I have played around with the HTML query and I absolutely love the result. Finally I can have the tag name and the angle brackets highlighted, but not the attributes. I have literally wanted this for years, ever since I first learned about rainbow parentheses. Thank you so much for your work!

I have also liked rainbow brackets for a long time and got used to it from emacs. This is the best plugin for it that I have seen in Neovim, so I am happy that we are able to improve on it too! (I was annoyed by Lua blocks highlighting in stead of html, but I also think that got much better with this update.)

HiPhish:
I get `! [remote rejected] Danielkonge/master -> Danielkonge/master (permission denied)` when I try to push to your remote (github.com:Danielkonge/rainbow-delimiters.nvim.git) on the `master` branch. Maybe you should have created a topic branch first?

HiPhish:
OK, looks like I figured it out. I had to push the commit as `git push Danielkonge-contrib HEAD:master` where `Danielkonge-contrib` is the remote name I have chose for your repo. This will make collaboration much easier.

lua/rainbow-delimiters/lib.lua
```diff
@@ -64,7 +64,7 @@ M.buffers = {}
 ---Fetches the query object for the given language from the settings.
 ---
 ---@param lang string  Name of the language to get the query for
----@return userdata query  The query object
+---@return Query|nil query  The query object
```
Danielkonge, line 67:
```suggestion
---@return Query? query  The query object
```
I can edit this later, but didn't want to do it while you might be editing. Just for consistency, this should be with `?` instead of `|nil`.

lua/rainbow-delimiters/strategy/global.lua
```diff
@@ -64,36 +75,63 @@ local function update_range(bufnr, changes, tree, lang)
...
-			for _, other in matches:iter() do
-				if not ts.is_ancestor(match_record.container, other.container) then
-					break
-				end
-				match_record.children:push(other)
-				matches:pop()
-			end
-			matches:push(match_record)
 		end
+		-- note: we have a dangling empty match_record that doesn't matter
```
Danielkonge, line 134:
```suggestion
```
This is no longer true.

lua/rainbow-delimiters/strategy/local.lua
Danielkonge, line 1: When you look at the code, consider updating line 175 in the local strategy. Stylua is complaining that iter_matches gets a wrong input.

queries/fennel/rainbow-delimiters.scm
```diff
@@ -1,31 +1,31 @@
...
-  ")" @closing) @container
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
 
 (table
-  "{" @opening
-  (":" @intermediate)*
-  "}" @closing) @container
+  "{" @delimiter
+  (":" @delimiter)*  ; FIXME: Only the first one is highlighted
```
Danielkonge, line 7:
```suggestion
  (":" @delimiter _)*  ; FIXME: Only the first one is highlighted
```
This seems to fix it for me, but I am not sure if we generally know that the type of `_` is `symbol` (that is what it is in the example code). If so, replace `_` with `(symbol)`.

queries/sql/rainbow-delimiters.scm
```diff
@@ -1,24 +1,24 @@
...
 ;; A parenthesized expression; the delimiters are actually outside the
 ;; expression, but that's OK because the nesting level only looks at containers
 (
-    "(" @opening
+    "(" @delimiter
     .
     [
         (binary_expression)
         (literal)
     ] @container
```
Danielkonge, line 21:
```suggestion
    ]
```
This `@container` should be moved to the end. But even with that fix, sql highlighting seems pretty broken in the example file.

Danielkonge: I added a fix to this now. It works perfectly for me as long as you don't have more than 3 layers of (((()))). I.e., ((())) seems to be ok.

lua/rainbow-delimiters/strategy/local.lua
Danielkonge: Update: I changed it from 
```lua
for _, match in query:iter_matches(tree:root(), bufnr) do
```
to 
```lua
for _, match in query:iter_matches(tree:root(), bufnr, 0, -1) do
```
now. I am pretty sure that was the behavior we were getting anyways.

queries/html/rainbow-delimiters.scm
```diff
@@ -1,19 +1,47 @@
...
+    (tag_name) @delimiter
+    "/>" @delimiter @sentinel)) @container
+
+;; The slash in a self-closing tag is optional
+;; This pattern does not actually work, it conflicts with first pattern.
+; (element
+;   (start_tag
+;     "<" @delimiter
+;     (tag_name) @delimiter
+;     ">" @delimiter @sentinel)) @container
```
Danielkonge, line 26: If we really want the self-closing tags without writing the `/`, then we can do something like this:
```suggestion
(element
  (start_tag
    "<" @delimiter
    (tag_name) @delimiter @ignore
    ">" @delimiter @sentinel)
  (#any-of? @ignore
   "area"
   "base"
   "br"
   "col"
   "embed"
   "hr"
   "img"
   "input"
   "link"
   "meta"
   "param"
   "source"
   "track"
   "wbr")
) @container
```

Danielkonge (commented):
We could add a check for `name == 'ignore'` or similar that does noting to the code, if we want to allow `#eq?` or similar in queries without logging it as an unknown capture name. See e.g. the following comment.

HiPhish:
SQL parentheses look fine to be as long as there are no empty parentheses or parentheses containing only one element.

Screenshot_20231028_003155

In a an expression line `SELECT (1 + (2));` all the characters inside the outer parentheses are on the same level in the tree.

```
      term [0, 7] - [0, 16]
        value: "(" [0, 7] - [0, 8]
        value: binary_expression [0, 8] - [0, 15]
          left: literal [0, 8] - [0, 9]
          operator: "+" [0, 10] - [0, 11]
          right: "(" [0, 12] - [0, 13]
          right: literal [0, 13] - [0, 14]
          right: ")" [0, 14] - [0, 15]
        value: ")" [0, 15] - [0, 16]
```

I don't think there is anything we can do without a container node.

Danielkonge:
> SQL parentheses look fine to be as long as there are no empty parentheses or parentheses containing only one element.

You mean after my update right? Because I was definitely having problems other than with nesting before that.

HiPhish:
Yes, after the update.

> Also, since writing something like `((()))` is very uncommon outside tests like this, I think it might actually make sense to simplify the code a bit and just let it look bad on something like that.

Agreed. In SQLite at least `SELECT (((1)));` is legal query, but no one is ever going to write code like this.

queries/html/rainbow-delimiters.scm
HiPhish: I like this. It will only work for select tag names, but it's a reasonably-sized list and it is easy to extend.

queries/fennel/rainbow-delimiters.scm
HiPhish: Thanks, that's a great trick. It will come in handy for other languages as well.

lua/rainbow-delimiters/strategy/local.lua
HiPhish: I want to leave cleaning up the local strategy for last. It is mostly the same as the global strategy, so one everything else is done I should have a better understanding of the new code.

Danielkonge:
I updated the local and global strategy so that any capture name that starts with 'ignore' will be ignored (we don't want to log those). So we can use `ignore`, `ignore1`, `ignore_self_closing`, etc. for any captures we want to use for predicates like `#eq?`, `#any-eq?`, etc.

lua/rainbow-delimiters/strategy/global.lua
```diff
@@ -123,13 +130,17 @@ local function update_range(bufnr, changes, tree, lang)
 					matches:push(match_record)
 					match_record = nil
 				end
+			elseif starts_with_ignore(name) and match_record then
+				-- do nothing
+				goto continue
```
HiPhish, line 135: `goto` is not part of Lua 5.1, it was added as an extension from a later version to LuaJIT. I know that *most* people will build Neovim with LuaJIT, but it is not guaranteed to be available on all platforms and in all distros. Do we really need a `goto` here? It looks to me like we would naturally get to the end of the if/else block.

```suggestion
```

Danielkonge: We don't need goto at all, but if it is just the comment, treesitter doesn't parse it nicely. I have changed it to an empty `do end` block now. I assume that should basically be equivalent to just a comment anyways.

queries/elixir/rainbow-delimiters.scm
```diff
@@ -1,31 +1,32 @@
...
+    ")" @delimiter @sentinel) @container)
 
 (block
-  "(" @opening
-  ")" @closing) @container
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
 
 ;; This pattern breaks the parser.
 ; Error attaching strategy to buffer 1: /usr/share/nvim/runtime/lua/vim/treesitter/query.lua:259: query: invalid structure at position 170 for language elixir
```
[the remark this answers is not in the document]
HiPhish: You are right. I have removed the note.

HiPhish:
> I updated the local and global strategy so that any capture name that starts with 'ignore' will be ignored (we don't want to log those). So we can use `ignore`, `ignore1`, `ignore_self_closing`, etc. for any captures we want to use for predicates like `#eq?`, `#any-eq?`, etc.

Why not just ignore any capture that has a name other than `container`, `delimiter` and `sentinel`? Currently unknown capture names are logged, but the function will run on every change and spam the logs. If it is about validating queries for query authors, we can do so only once when the query is loaded first like this:

```lua
local known_captures = {delimiter = true, container = true, sentinel = true}

local q = vim.treesitter.query.get('lua', 'rainbow-delimiters')
for _, capture in ipairs(q.captures) do
    if not known_captures[capture] or not starts_with_ignore(capture) then
        log.error('Unknown capture name: ' .. capture)
    end
end
```

This validation can be added somewhere in the `lib` module, for example when registering a new namespace. Then the validation will only run once over the entire lifetime of the Neovim process.

I would not add validation to the scope of this PR, it is big enough as it is. We can open a new ticket for that.

Danielkonge:
> Why not just ignore any capture that has a name other than `container`, `delimiter` and `sentinel`? Currently unknown capture names are logged, but the function will run on every change and spam the logs. If it is about validating queries for query authors, we can do so only once when the query is loaded first like this:
>
> ```lua
> local known_captures = {delimiter = true, container = true, sentinel = true}
> 
> local q = vim.treesitter.query.get('lua', 'rainbow-delimiters')
> for _, capture in ipairs(q.captures) do
>     if not known_captures[capture] or not starts_with_ignore(capture) then
>         log.error('Unknown capture name: ' .. capture)
>     end
> end
> ```
> 
> This validation can be added somewhere in the `lib` module, for example when registering a new namespace. Then the validation will only run once over the entire lifetime of the Neovim process.
> 
> I would not add validation to the scope of this PR, it is big enough as it is. We can open a new ticket for that.

I think you are right, so I have removed the unknown capture name cases now. I still have the case where we get a `@delimiter` or a `@sentinel` without having a `match_record`. As long as the queries (and code) are correct we should never end up in this case, so it shouldn't be spamming the logs. Do you want to change this case too?

HiPhish:
> Do you want to change this case too?

No, please leave it in. This is an error case which cannot be detected until the code actually runs, so having the error in the logs is valuable.

lua/rainbow-delimiters/strategy/global.lua
```diff
@@ -123,30 +158,41 @@ local function setup_parser(bufnr, parser)
...
+				-- otherwise we use the whole tree's range
+				if p._parent_lang then
+					-- HACK: changes can accidentally overwrite highlighting in
+					-- injected code blocks
+
+					-- We have a parent language, so we are in an injected
+					-- language code block, thus we update all of the current
+					-- code block
+					changes = {{tree:root():range()}}
+				else
```
HiPhish, line 176: Can you please elaborate on why you are using the full tree here?

Let's say I have Lua code with embedded Vim script:

```lua
print('Hello from lua')
vim.cmd [[
echo 'Hello from Vim'
echo 'tests'
echo 2 + 3 == 5
echo 'Goodbye from Vim'
]]
print('Goodbye from lua')
```

What is the scenario this is trying to solve?

Danielkonge: The scenario this tries to solve doesn't seem to show up in Lua code with embedded Vim script (presumably the treesitter implementation catching the changes is better there), but it does show up in markdown with injected languages.

E.g., if you have something like (but with a proper injected language block -- I am not sure how to nest them in markdown):

```markdown
Some markdown. 
`vim
echo 'Hello from Vim'
echo 'tests'
`
More markdown.
```
Then if you move the top line out of the injected language block at the top and move it back down, you get correct looking changes for the `vim` block, but the `markdown` changes are registered as the whole `vim` block changing. Thus Vim highlighting will be reset in the block (by the markdown changes), and then only updated on the first line (by the `vim` block changes). 

This is a pretty rare problem, but it show up in my testing with markdown.

HiPhish: Does it still happen with the current code? I tried to replicate your example; first I executed `:move 1` while my cursor was on line 3, then I executed `:move 3` while my cursor was on line 2 to move the text back into the block. The result looks exactly as before. The changes the firs time around are

- markdown `{ { 0, 14, 1, 21 } }`
- vim `{}`

The second time around (moving the line back) the changes are

- markdown `{ { 0, 14, 1, 0 }, { 2, 0, 4, 0 } }`
- vim `{ { 2, 0, 3, 0 } }`

Could it be that the order is the other way around for you? I am running Neovim 0.9.4 on GNU/Linux.

HiPhish:
I have noticed a problem with nested languages: when I move a line up out of a nested block it will retain its rainbow highlighting. Here is a Markdown example

```markdown
Some markdown.
~~~lua
print({{{{}}}})
print({{{{}}}})
~~~
More markdown
```

I place the cursor on line 3 and execute `:move 1`. The line will lose its Lua highlighting, but it will retain its rainbow highlighting. However, if I place the cursor on line 4 and execute `:move 5` (i.e. downwards) it will lose the rainbow highlighting.

Screenshot_20231029_233710

lua/rainbow-delimiters/strategy/global.lua
Danielkonge: Here is the behavior for me with the current code:

https://github.com/HiPhish/rainbow-delimiters.nvim/assets/7075380/1b8025d8-9b77-47e4-996d-aa0acd279736

and with the entire `if`/`else` block commented out but 
```lua
changes = vim.tbl_map(normalize_change, changes)
```
added outside the `if`/`else` block instead:


https://github.com/HiPhish/rainbow-delimiters.nvim/assets/7075380/66b84842-dab0-4c25-9e54-1773ccb0aa9d


It might be because we move text differently? I move text with these keymaps:
```lua
-- keymaps for moving marked text up or down
vim.keymap.set("v", "J", ":m '>+1<CR>gv=gv", { silent = true, desc = "Move marked text down" })
vim.keymap.set("v", "K", ":m '<-2<CR>gv=gv", { silent = true, desc = "Move marked text up" })
```

Danielkonge:
> I have noticed a problem with nested languages: when I move a line up out of a nested block it will retain its rainbow highlighting. Here is a Markdown example
> 
> ```
> Some markdown.
> ~~~lua
> print({{{{}}}})
> print({{{{}}}})
> ~~~
> More markdown
> ```
> 
> I place the cursor on line 3 and execute `:move 1`. The line will lose its Lua highlighting, but it will retain its rainbow highlighting. However, if I place the cursor on line 4 and execute `:move 5` (i.e. downwards) it will lose the rainbow highlighting.
> 
> Screenshot_20231029_233710

I can reproduce this if I copy exactly your short markdown file there, but it is not a problem in the stress markdown file, so I am not sure what causes this. I will try to look at the changes returned some more.

lua/rainbow-delimiters/strategy/global.lua
Danielkonge: See my comment below. I think it might actually just be the different files behaving differently and not the commands.

Danielkonge:
I think I figured out the problem. It seems to be an off by one error from what I can tell. 

Changing 

```lua
vim.api.nvim_buf_clear_namespace(bufnr, nsid, change[1], change[3])
```

to

```lua
vim.api.nvim_buf_clear_namespace(bufnr, nsid, change[1], change[3] + 1)
```

fixes the problem. 

https://github.com/HiPhish/rainbow-delimiters.nvim/assets/7075380/bfd1df5e-2ba9-4559-ace4-a6e1cb9d2136


But this then breaks `html` highlighting. I originally had the lower one, but it gives this highlighting:

Screenshot 2023-10-30 at 00 13 40

instead of:

Screenshot 2023-10-30 at 00 14 08

It seems to be a problem with clearing out the old highlights caused by different behavior in the different parsers. I am not yet sure though, so I will try to look into this a bit more.

Danielkonge:
https://github.com/HiPhish/rainbow-delimiters.nvim/pull/43/commits/a8ff2a79ec95c5894d98135dd2a2bc68f14f25c6

This is a temporary fix that seems to work in almost all cases I have checked. The only way I have managed to break this is by having nested language injections (i.e., a language injection in a language injection), but that seems like a rare need. In those rare cases, I think it makes sense to refresh highlighting (that will fix it) if you move code.

Danielkonge:
I cleaned up the code a bit and noticed another problem with injections in Rust. When you use macros in Rust, it treats that as a language injection with the language being Rust, but when we inject the language itself, it will already be taken care of by the part of the code without a parent language, so in that case we should do nothing. This has been messing with Rust highlighting of (and around) macros for me for a long time, and the current fix I made finally gets around it. 

I have tested out the code some more and haven't noticed any problems now except when you have nested language injections, which I think is pretty rare. In those cases updating highlighting or moving stuff can break the highlighting with the current code. E.g.:
```markdown
Markdown text.
~~~lua
print 'This is injected Lua in Markdown (one level deep)'
vim.cmd[[
echo str([])
echo 'This is injected Vim in Lua in Markdown (two levels deep)'
]]
~~~
```

HiPhish:
You have the patience of a saint. It will be a day or two before I get around to looking at the code, so I need to ask for a bit more patience.

Danielkonge:
Instead of `ignore<anything>` for the captures that we ignore for both debug and highlighting, do you think it would make more sense to use `_<anything>`? I.e., just ignore stuff starting with underscore. I feel like that might be a bit more expected behavior.

HiPhish:
Yes, underscore is better. It is more in line with ignored variables in programming languages and harder to mistype.

Danielkonge:
In https://github.com/HiPhish/rainbow-delimiters.nvim/pull/43/commits/d4cc189f80dedf327aa2308033823cb515a830d1 I included some comments that I will delete soon, but I made a commit to look back on if needed.

I played around with using `vim.api.nvim_get_extmarks` and `vim.api.nvim_del_extmark` instead of `vim.api.nvim_buf_clear_namespace` since those allow us to work with exact rows and columns instead of just the rows, but I didn't manage to improve the behavior over what we have now (there was still a problem with nested injections), so I didn't switch to those functions in the code.

HiPhish:
I just reordered some code to remove one level of if-nesting. Can you please double-check that the conditions are still the same?

Danielkonge:
> I just reordered some code to remove one level of if-nesting. Can you please double-check that the conditions are still the same?

The logic looks correct to me. (Also, it behaves as expected in quick testing of what that code is supposed to solve.)

HiPhish:
Where did you get the `_parent_lang` field of the parser from? It is not documented, and when I inspect a parser object there is no such field either. I used the following Markdown text:

~~~md
Hello

```c
puts((((("This is an injected language")))));
{
    {
        {
            {
                {
                    return ((((((((((((2)))))))))))) + ((((((3))))));
                }
            }
        }
    }
}
```
~~~

Then I set `p = vim.treesitter.get_parser()` and executed `lua =p` to view the fields of the parser object. Each parser has a `children` field which maps contained languages to their respective parsers. But there is no `_parent_lang` field.

Danielkonge:
> Where did you get the `_parent_lang` field of the parser from? It is not documented, and when I inspect a parser object there is no such field either. I used the following Markdown text:
> 
> ```
> Hello
> 
> ```c
> puts((((("This is an injected language")))));
> {
>     {
>         {
>             {
>                 {
>                     return ((((((((((((2)))))))))))) + ((((((3))))));
>                 }
>             }
>         }
>     }
> }
> ```
> ```
> 
> Then I set `p = vim.treesitter.get_parser()` and executed `lua =p` to view the fields of the parser object. Each parser has a `children` field which maps contained languages to their respective parsers. But there is no `_parent_lang` field.
> 
> EDIT: I don't see any way of finding out whether the parser has a parent language. We can descend deeper into the tree, but not back up.

I found `_parent_lang` by printing `p` in `on_changedtree` and then going through `:messages`. It seems to be `nil` (so not shown) when there is no parent language, but it should show up for all injections. 

Screenshot 2023-11-02 at 00 19 45

Screenshot 2023-11-02 at 00 20 28

Screenshot 2023-11-02 at 00 21 05

If we want to avoid using `_parent_lang` and still have movement of text work with injections, I think we need to figure out what causes the wrong `changes` to be reported in certain cases (like the above c code). If that is fixed, then we could simplify the code a bit and even have it work with nested injections.

Danielkonge:
Since `p._parent_lang` seems to be a nightly feature, I have updated `util.for_each_child` so that we can keep track of the parent language ourselves instead. Everything seems to work as expected in the global strategy now (I haven't made the corresponding changes in the local strategy yet). 

Please let me know if there is any problems with this update.

Danielkonge:
I think I have fixed the problem with nested injections now, but the solution is a bit hacky. I don't think it should affect the performance of the highlighting though, since the main part of the updated code will only be run when we are at least two injections deep. 

Please let me know if you notice any problems. 

Note: If we don't want to include this last change, the previous code before this change worked for everything other than nested injections in my testing.

HiPhish:
I don't think we need the function `for_each_child_dif_lang` if we move the test into the thunk. I have done so in the previous commit, please check if the logic is the same.

HiPhish:
Your changes now work for me as well. One more questions:

What does this mean? Can you please comment it?

```
				if changes[1] and changes[1][1] then
					update_range(bufnr, changes, tree, lang)
				end
```

The first test checks that `changes` is not an empty array, but why the second one? The items of `changes` are four-element range tables, so how can `changes[1][1]` be `nil`?

HiPhish:
> I think I have fixed the problem with nested injections now, but the solution is a bit hacky. I don't think it should affect the performance of the highlighting though, since the main part of the updated code will only be run when we are at least two injections deep.

I think it's fine, albeit quite ugly because you have to drag information about two languages with you. Maybe I can find a nicer way, but if not I'd leave it as it is. It may be ugly, but it works.

Danielkonge:
> I don't think we need the function `for_each_child_dif_lang` if we move the test into the thunk. I have done so in the previous commit, please check if the logic is the same.

I was about to say that this might cause problems, but looking at it again it seems that the `return` just applies to the inner thunk function, so it should work as expected, yes.

Though, to clarify the type of the thunk function I changed it from `function(child_parser)` to `function(child_parser, _, _)`.

Danielkonge:
> Your changes now work for me as well. One more questions:
> 
> What does this mean? Can you please comment it?
> 
> ```
> 				if changes[1] and changes[1][1] then
> 					update_range(bufnr, changes, tree, lang)
> 				end
> ```
> 
> The first test checks that `changes` is not an empty array, but why the second one? The items of `changes` are four-element range tables, so how can `changes[1][1]` be `nil`?

I don't think there is any reason to check `changes[1][1]` in the current code, so I removed it. 

I think at some point I might have had something like `{{}}` for changes, which is why I had that check.

Danielkonge:
> > I think I have fixed the problem with nested injections now, but the solution is a bit hacky. I don't think it should affect the performance of the highlighting though, since the main part of the updated code will only be run when we are at least two injections deep.
> 
> I think it's fine, albeit quite ugly because you have to drag information about two languages with you. Maybe I can find a nicer way, but if not I'd leave it as it is. It may be ugly, but it works.

If we could get an upstream update that reported changes correctly (or at least what I would consider to be the correct way) all the time instead of sometimes reporting the whole injection block for some injected languages, then it should be possible to simplify the code quite a bit.

HiPhish:
In this block:

```lua
				-- If a line has been moved from another region it will still carry with it
				-- the extmarks from the old region.  We need to clear all extmarks which
				-- do not belong to the current language
				for _, change in ipairs(changes) do
					if change[1] then
						for key, nsid in pairs(lib.nsids) do
							if key ~= lang then
								-- HACK: changes in the main language sometimes need to overwrite
								-- highlighting on one more line
								local line_end = change[3] + (parent_lang and 0 or 1)
								vim.api.nvim_buf_clear_namespace(bufnr, nsid, change[1], line_end)
							end
						end
					end
				end
```

the check `if change[1] then` seems superfluous as well to me. Am I missing something?

HiPhish:
Let's see if I understand the problem correctly. Let's say I have this Markdown code:

~~~markdown
Hello

```c
puts((((("This is an injected language")))));




puts((((("This is an injected language")))));
```
~~~

If I place the cursor on line 4 and execute `:move 1` I get the following changes:

- `markdown { { 0, 5, 5, 1, 45, 51 } }`
- `c { { 4, 0, 58, 8, 0, 62 } }`

This is fine and exactly what I would expect. Now suppose I place my cursor on line 2 (containing the moved text) and execute `:move 4` to move it back into the C block. Then I get the following changes:

- `markdown { { 0, 5, 5, 1, 0, 6 }, { 3, 0, 12, 9, 0, 108 } }`
- `c { { 3, 0, 12, 8, 0, 62 } }`

The first Markdown change encompasses the first two lines, which is fine. However, the second Markdown change encompasses lines 4 to 10, which contain C code. Thus the extmarks in the C block get erased. Then the changes from the C block (line 4 to 9) are applied. But this should restore the extmarks again, should it not?

What is the faulty behaviour you are observing?

Danielkonge:
> Let's see if I understand the problem correctly. Let's say I have this Markdown code:
> 
> ~~~
> Hello
> 
> ```c
> puts((((("This is an injected language")))));
> 
> 
> 
> 
> puts((((("This is an injected language")))));
> ```
> ~~~
> 
> If I place the cursor on line 4 and execute `:move 1` I get the following changes:
> 
> * `markdown { { 0, 5, 5, 1, 45, 51 } }`
> * `c { { 4, 0, 58, 8, 0, 62 } }`
> 
> This is fine and exactly what I would expect. Now suppose I place my cursor on line 2 (containing the moved text) and execute `:move 4` to move it back into the C block. Then I get the following changes:
> 
> * `markdown { { 0, 5, 5, 1, 0, 6 }, { 3, 0, 12, 9, 0, 108 } }`
> * `c { { 3, 0, 12, 8, 0, 62 } }`
> 
> The first Markdown change encompasses the first two lines, which is fine. However, the second Markdown change encompasses lines 4 to 10, which contain C code. Thus the extmarks in the C block get erased. Then the changes from the C block (line 4 to 9) are applied. But this should restore the extmarks again, should it not?
> 
> What is the faulty behaviour you are observing?

Your changes look fine, but I don't always get changes like that.

In this block in the stress markdown file:
~~~markdown
Fata quae: **falsi**: pares dea [agmine
hospite](http://www.quo-retro.net/sublimis) catenas? [Morte](http://corpus.net/)
ad putes. Dicentem Chariclo vidit!

```c
puts("This is an injected language")
{
    {
        {
            {
                {
                    return ((((((2)))))) + ((((3))))
                }
            }
        }
    }
}
```
~~~

If I move the line with `puts` up one line and then down one line again, I get the following changes:
```
markdown
{ { 42, 0, 1123, 43, 0, 1160 } }
c
{}
markdown
{ { 43, 0, 1128, 55, 0, 1318 } }
c
{ { 43, 0, 1128, 44, 0, 1165 } }
```

The first markdown and c changes are fine, but the second markdown changes cover the whole c block, so it clears the highlighting there. And then the c changes only (correctly) catches the change on the first line of the C block. I.e., I think the changes reported by the second markdown block is wrong.

Danielkonge:
> In this block:
> 
> ```lua
> -- If a line has been moved from another region it will still carry with it
> -- the extmarks from the old region.  We need to clear all extmarks which
> -- do not belong to the current language
> for _, change in ipairs(changes) do
> 	if change[1] then
> 		for key, nsid in pairs(lib.nsids) do
> 			if key ~= lang then
> 				-- HACK: changes in the main language sometimes need to overwrite
> 				-- highlighting on one more line
> 				local line_end = change[3] + (parent_lang and 0 or 1)
> 				vim.api.nvim_buf_clear_namespace(bufnr, nsid, change[1], line_end)
> 			end
> 		end
> 	end
> end
> ```
> 
> the check `if change[1] then` seems superfluous as well to me. Am I missing something?

With the current code, I think you are right, I have removed it now.

HiPhish:
I can reproduce the issue with the changes. I's say this wraps up things for the global strategy. If you have any changes to make to the local strategy let me know, otherwise I will start reviewing the code there. Once the local strategy is done I think we can finally finish all of this and merge it into `master`.

In the meantime I will go ahead and document the problem with the changes.  I am not sure if this is really an issue or just a really weird implementation quirk. I mean, highlighting should be running into the same problem as well, but somehow they are avoiding the issue.

HiPhish (approved):

Danielkonge:
> I can reproduce the issue with the changes. I's say this wraps up things for the global strategy. If you have any changes to make to the local strategy let me know, otherwise I will start reviewing the code there. Once the local strategy is done I think we can finally finish all of this and merge it into `master`.
> 
> In the meantime I will go ahead and document the problem with the changes. I am not sure if this is really an issue or just a really weird implementation quirk. I mean, highlighting should be running into the same problem as well, but somehow they are avoiding the issue.

I have already updated the local strategy so it builds the match tree similarly to how the global strategy works, and based on my testing it works as expected now. I think it should be ready for review. Then we can make larger changes to it in another pull request later, if there is something that could improve the way it works in general. 

Also: I noticed that we never actually used `match_record.sentinel` in the code now (since we use information about the `match_record` being nil or not instead now), so I removed that from the code.

Danielkonge:
I have fixed the local strategy so it should work with multiple injections of the same language now (this didn't work before based on my testing). I think there are other changes that we could make, but before doing that it is probably worth it to reconsider how the local strategy should work in more detail instead.

In other words, I think the local strategy code is ready for review now, and then we can consider changing the local strategy more later.

HiPhish:
In the local strategy, what is the point of adding another level to `match_trees`? My guess is that it is for when there are multiple places where another language is injected. For example a Markdown buffer might contain multiple Lua code blocks and we need a way of distinguishing the different trees. Is that correct?

Danielkonge:
> In the local strategy, what is the point of adding another level to `match_trees`? My guess is that it is for when there are multiple places where another language is injected. For example a Markdown buffer might contain multiple Lua code blocks and we need a way of distinguishing the different trees. Is that correct?

That is correct, yes. The old code just used the last code block of a given injected language, so it wouldn't highlight any other block of that language properly. 

Also, a quick question: You write `-- NOTE: according to the Lua 5.1 manual `%p` is not supported, yet it works`, but I thought it was included: https://www.lua.org/manual/5.1/manual.html#5.4.1

Are there multiple different manuals? Or do you mean that `%p` has another meaning according to the manual? (punctuation)

HiPhish:
> Also, a quick question: You write `-- NOTE: according to the Lua 5.1 manual `%p` is not supported, yet it works`, but I thought it was included: https://www.lua.org/manual/5.1/manual.html#5.4.1
> 
> Are there multiple different manuals? Or do you mean that `%p` has another meaning according to the manual? (punctuation)

The Lua 5.1 manual is in `:h luaref`. You have linked to the regex pattern, which are different from string format options (they just happen to look the same). Under `:h string.format` it says the following:

> The format string follows the same rules as the `printf` family of standard C functions. The only differences are that the options/modifiers `*`, `l`, `L`, `n`, `p`, and `h` are not supported and that there is an extra option, `q`.

In C the `%p` prints the memory address of a pointer. This is also what the `%p` in Neovim does. My guess is that it is provided as a non-standard extension by LuaJIT. On platforms where LuaJIT is supported that is most likely what will be running, but LuaJIT is not guaranteed.

But do we really need the memory address in the first place? Could we just use the table itself as the key? If we use a table as the key then only this particular unique table will work, which should have the same effect as using the address.

Danielkonge:
> > Also, a quick question: You write `-- NOTE: according to the Lua 5.1 manual `%p` is not supported, yet it works`, but I thought it was included: https://www.lua.org/manual/5.1/manual.html#5.4.1
> > Are there multiple different manuals? Or do you mean that `%p` has another meaning according to the manual? (punctuation)
> 
> The Lua 5.1 manual is in `:h luaref`. You have linked to the regex pattern, which are different from string format options (they just happen to look the same). Under `:h string.format` it says the following:
> 
> > The format string follows the same rules as the `printf` family of standard C functions. The only differences are that the options/modifiers `*`, `l`, `L`, `n`, `p`, and `h` are not supported and that there is an extra option, `q`.
> 
> In C the `%p` prints the memory address of a pointer. This is also what the `%p` in Neovim does. My guess is that it is provided as a non-standard extension by LuaJIT. On platforms where LuaJIT is supported that is most likely what will be running, but LuaJIT is not guaranteed.
> 
> But do we really need the memory address in the first place? Could we just use the table itself as the key? If we use a table as the key then only this particular unique table will work, which should have the same effect as using the address.

I did try to just use the table at first, but I ran into some problems. Maybe I made a mistake somewhere else (I remember updating other parts of the code), I will try to see if I can change it and make it work with just tables now.

HiPhish:
Alternatively you can apply `tostring` to the table. The string will always have the prefix `table `, but that shouldn't matter.

Danielkonge:
I changed it now and just using the table as the key works fine. I must have edited something else while introducing the `get_id` function that fixed the actual problem. (I think it was the `match_trees[bufnr][sub_lang] = {}` code.)

HiPhish:
I think I am done now. It's getting very late in my time zone, so I don't want to start cleaning up the Git history now. I propose that tomorrow I rebase the PR into a few nice and clean commits and then we leave it as it for a day or two in case we find more edge cases.

Unrelated question: when you add type annotations like `TSTree` the Lua language server gives me a warning about the unknown type. Have you figured out how to teach the server about these custom Neovim types, or do you just live with the warning?

Danielkonge:
> I think I am done now. It's getting very late in my time zone, so I don't want to start cleaning up the Git history now. I propose that tomorrow I rebase the PR into a few nice and clean commits and then we leave it as it for a day or two in case we find more edge cases.

That sounds like a good idea.

> Unrelated question: when you add type annotations like `TSTree` the Lua language server gives me a warning about the unknown type. Have you figured out how to teach the server about these custom Neovim types, or do you just live with the warning?

LuaLS suggested that type to me itself and I don't get that warning. I am pretty sure that it is [neodev](https://github.com/folke/neodev.nvim) giving me those annotations (I don't think I have anything else special in my Lua LSP setup other than `Lua.diagnostics.unusedLocalExclude = { "_*" }` to allow `_name` to be unused without a warning). 

This is basically my current LSP setup (other than a lot of keymaps):
Screenshot 2023-11-05 at 02 05 02

HiPhish:
I have squashed the MR down to three commits. Unfortunately this also changed the author to myself. If you want to get credit for what has been mostly your work you can edit the commit messages to add a `Co-authored-by` trailer: https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors

HiPhish:
> I am pretty sure that it is [neodev](https://github.com/folke/neodev.nvim) giving me those annotations

Yes, that was it, thank you.

Danielkonge:
> I have squashed the MR down to three commits. Unfortunately this also changed the author to myself. If you want to get credit for what has been mostly your work you can edit the commit messages to add a `Co-authored-by` trailer: https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors

I added myself as a co-author now.

Danielkonge:
Do you prefer not to write vimdoc language injections like
```vimdoc
>lua
if something then
  return true
else
  return false
end
<
```
similar to #46?

Otherwise, I think adding those + adding an empty file under `vimdoc/rainbow-delimiter.scm` in `queries` is a good idea. This makes it so that the injected languages will also use rainbow delimiters for highlighting, which I think makes a lot of sense for the docs in this case.

Note: I would just add an empty file since the `vimdoc` treesitter tree doesn't really allow us to write good queries for delimiters (other than `{`, `}` pairs which are already highlighted nicely).

HiPhish:
I did not even know that `>lang` was a thing until a few days ago. I am definitely in favour of supporting it in the manual, just not for this PR. Let's get this out of the door first.

Anyway, I have added an empty file now. I also forgot to pull your changes, so you will have to re-add yourself for credit again. Sorry about that.

> Also, we should probably update the version number since this breaks older queries.

I have not been versioning this plugin since everything has been in flux. But I think this is a good point to start finally versioning it.

Danielkonge:
> I have not been versioning this plugin since everything has been in flux. But I think this is a good point to start finally versioning it.

Do you want to add a version number in the docs (you have it right at the beginning of `doc/rainbow-delimiters.txt`) before I redo the co-author notes on the commits? 

(I assume the changelog can just be added later after this pull request, but we might as well add the correct version in the docs for when this is committed on master.)

HiPhish:
I prefer to add version numbers directly to `master`. My workflow is to have a `master` branch that is always ready for use and tag specific milestone commits as versions. Features are developed in feature branches, then merged into `master` once they are ready for use. When I am satisfied with the feature set I tag a new release on `master`.

I admit it is somewhat arbitrary in this plugins what constitutes a release. Like, if I add a new pattern to an existing query, does this warrant a new release? It kind of does, even though the change is trivial. Usually I let it sit for a few days in case something comes up before I tag a version.

Danielkonge:
> I prefer to add version numbers directly to `master`. My workflow is to have a `master` branch that is always ready for use and tag specific milestone commits as versions. Features are developed in feature branches, then merged into `master` once they are ready for use. When I am satisfied with the feature set I tag a new release on `master`.

That makes sense. You can just update the version number later, when this is on your master branch then. 

> I admit it is somewhat arbitrary in this plugins what constitutes a release. Like, if I add a new pattern to an existing query, does this warrant a new release? It kind of does, even though the change is trivial. Usually I let it sit for a few days in case something comes up before I tag a version.

For this plugin, I think it would make sense to update `z` in version `x.y.z` for a collection of smaller query update commits or each query update commit (depending on your preferences), and then update either `x` or `y` for something like this pull request (again, depending on your preferences, though I would probably just update `y`). I don't think the exact versioning choice is that important, it should just be clear that there is a version change with breaking changes like this. But, as mentioned, you can easily do that later when you commit this to the main branch.

Also, I have added myself as a coauthor again.

HiPhish:
@Danielkonge I have been daily driving this for the last couple of days and as far as I am concerned we can merge it into `master`. Any objections from your side?

Danielkonge:
> @Danielkonge I have been daily driving this for the last couple of days and as far as I am concerned we can merge it into `master`. Any objections from your side?

I have been using it too and haven't noticed any problems, so I think it is ready to be merged. So no objections from me.

HiPhish:
Thank you very much. It has been an honour and a pleasure working with you. This inability to capture more than one node has been a thorn in my side from the start and I don't think I would have come up with a solution anytime soon. Your persistence with all the edge cases for embedded languages was especially amazing.

Speaking of embedded languages, this is really something that would benefit from more rigorous integration testing. I have Vader tests, but Vader has the problem that it runs all tests in the same Vim instance, so tests can step on each other's toes. I have been thinking about re-implementing Vader for Neovim such that every test runs in its own Neovim instance (via `jobstart` in a pseudo terminal), but this would require Neovim 0.10 because that version ships with [lpeg](https://www.inf.puc-rio.br/~roberto/lpeg/). I really don't want to have to write a Vader parser by hand.

Danielkonge:
> Thank you very much. It has been an honour and a pleasure working with you. This inability to capture more than one node has been a thorn in my side from the start and I don't think I would have come up with a solution anytime soon. Your persistence with all the edge cases for embedded languages was especially amazing.

Thank you for such nice words, you have also been a pleasure to work with. 

> Speaking of embedded languages, this is really something that would benefit from more rigorous integration testing. I have Vader tests, but Vader has the problem that it runs all tests in the same Vim instance, so tests can step on each other's toes. I have been thinking about re-implementing Vader for Neovim such that every test runs in its own Neovim instance (via `jobstart` in a pseudo terminal), but this would require Neovim 0.10 because that version ships with [lpeg](https://www.inf.puc-rio.br/~roberto/lpeg/). I really don't want to have to write a Vader parser by hand.

Do you want to re-implement Vader to better support Vim and not just Neovim? Otherwise, I think the most common way I see other people do testing for Neovim plugins is by using [plenary](https://github.com/nvim-lua/plenary.nvim#plenarytest_harness). I assume you are already aware of plenary, so is there something you think it is missing?

HiPhish:
I have Plenary installed as a dependency for Telescope, but I never actually looked at what's inside. If it is good for integration testing I might as well use it instead of rolling my own. I do like the Vader file syntax, but I am not crazy about all the extra work of re-implementing it.

> Do you want to re-implement Vader to better support Vim and not just Neovim?

I wanted to focus on just Neovim, I haven't used Vim for over half a decade at least. I don't even know if Vim can be run inside a pseudo terminal and controlled from the outside like Neovim can.

PR #45, opened 2023-10-14
Add Janet support

sogaiu:
This is an attempt to add support for [Janet](https://janet-lang.org).

The test file (`regular.janet`) seemed to display fine here.

---

On a side note, the [contributing file](https://github.com/HiPhish/rainbow-delimiters.nvim/blob/652345bd1aa333f60c9cbb1259f77155786e5514/CONTRIBUTING.rst) currently mentions:

  `:h ts-rainbow-custom-queries`

May be the following is meant since the transition?

  `:h rb-delimiters-custom-query`

### DIFF
--- a/queries/janet_simple/rainbow-delimiters.scm
+++ b/queries/janet_simple/rainbow-delimiters.scm
@@ -1,0 +1,24 @@
+(par_tup_lit
+  "(" @opening
+  ")" @closing) @container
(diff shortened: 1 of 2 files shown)


HiPhish:
Thank you very much. And yes, you are right, I have now fixed the contributing file.

PR #46, opened 2023-10-15
Documentation improvements

thetic:
Apply syntax highlighting to code examples and explain purpose of `rainbow-delimiters.setup`.

### DIFF
--- a/doc/rainbow-delimiters.txt
+++ b/doc/rainbow-delimiters.txt
@@ -145,8 +145,8 @@ Alternatively, the same configuration in Lua:
                                                     *rainbow-delimiters.setup*
 'rainbow-delimiters.setup'.setup({config})
 
-Some people prefer to call a Lua `setup` function, so a setup function is
-available as part of a Lua module.
+While |g:rainbow_delimiters| is the recommended configuration method, a Lua
+`setup` function is available for compatibility with common plugin managers.
 >lua
     require 'rainbow-delimiters.setup'.setup {
         strategy = {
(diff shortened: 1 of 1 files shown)


doc/rainbow-delimiters.txt
```diff
@@ -145,9 +145,9 @@ Alternatively, the same configuration in Lua:
                                                     *rainbow-delimiters.setup*
 'rainbow-delimiters.setup'.setup({config})
 
-Some people prefer to call a Lua `setup` function, so a setup function is
-available as part of a Lua module.
->
+While |g:rainbow_delimiters| is the recommended configuration method, a Lua
+`setup` function is available for compatibility with common plugin managers.
+>lua
     require 'rainbow-delimiters.setup'.setup {
```
Danielkonge, line 151: I think there was a metatable update recently that simplified this, so you can just use `require 'rainbow-delimiters'.setup` (which will also work with e.g. Lazy). @HiPhish will know if this is true. 

If so, this could be simplified using that fact.

doc/rainbow-delimiters.txt
```diff
@@ -172,9 +172,6 @@ available as part of a Lua module.
 The keys are exactly the same as for |g:rainbow_parentheses|.  If fact, this
 function does the same as setting the variable directly.
 
-As an aside, this is a bad practice carried over from a time when Lua support
-in Neovim still had issues with Vim script interoperability, but it has
-persisted through cargo-culting.  You are better off not using this function.
```
Danielkonge, line 177: If this is what @HiPhish still thinks, then I don't think this should be removed.

doc/rainbow-delimiters.txt
```diff
@@ -145,9 +145,9 @@ Alternatively, the same configuration in Lua:
                                                     *rainbow-delimiters.setup*
 'rainbow-delimiters.setup'.setup({config})
```
Danielkonge, line 146: Same comment as below. Maybe this can be `'rainbow-delimiters'.setup({config})`?

Danielkonge (commented):
I had a few comments.

Danielkonge:
Note that the syntax highlighting has been added now, so unless you really want to reword some of the above in the docs, this pull request doesn't really change anything now.

PR #52, opened 2023-11-09
feat(elm): Add elm queries

nunntom:
Adds queries for [elm](https://elm-lang.org/).

### DIFF
--- a/queries/elm/rainbow-delimiters.scm
+++ b/queries/elm/rainbow-delimiters.scm
@@ -1,0 +1,63 @@
+(exposing_list
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
+
+(exposed_operator
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
+
+(exposed_union_constructors
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
+
+(_
+  "(" @delimiter
+  .
+  (type_expression)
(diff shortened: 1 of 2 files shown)


HiPhish:
Thanks a lot, but this PR is going to be obsolete soon because we have changed how queries are written in #43. Porting is straight-forward, I should be able to do it myself once the other PR is merged.

HiPhish:
I have rebased on the current `master` and ported all patterns except for `union_pattern`. The issue is that the `@container` node needs to contain the two `@delimiter` nodes, or else the colours won't be assigned properly. I don't know Elm, so I don't know how to solve this.

HiPhish:
Never mind, I think I have fixed `union_pattern` by wrapping it inside a `_` node which is the container of the whole pattern. Let me know if this is correct.

nunntom:
Hi. Thanks Union pattern seems to be working fine but type_expression is now not highlighting.

I think it's a similar issue. The delimiters are not inside the container and type_ref isn't working as a container.

image

nunntom:
It's working now by wrapping `type_expression` inside `_` as well.

HiPhish:
Looks good to me. Can you please squash all the commits into one? I will merge the PR then.

nunntom:
Done. Thanks!

HiPhish:
Merged, thank you.

Issue #53, opened 2023-11-10
[Bug]: Some comments cause neovim to slow down and freeze when present with this plugin


wrbright:
### Neovim version

0.9.4

### Language affected

Lua, vim

### Description

Text
When opening a file with less normal commenting, this plugin seems to slow the application to a crawl and eventually completely freeze Neovim.

Comments like the following effect the application.

```
" """"""""""""""""" "
" This is a comment "
" """"""""""""""""" "

----------
- This is another comment
----------
```

HiPhish:
Do you have the `comment` grammar installed? That parser will create a separate node for each `-` or `"` and each word inside the comment. Please try without the `comment` grammar first if you have it installed.

HiPhish:
Can you please check the current `master`? There has been a major update and it might have fixed the problem as a side effect.

> By my understanding at least three other projects were created by others which had ultimately been archived

To be honest, I can't blame them. All pre-Treesitter projects were doomed from the start, there is only so far you can get with regex hackery before the cracks start forming. [p00f/nvim-ts-rainbow](https://github.com/p00f/nvim-ts-rainbow) had potential, but the author was trying to do too much with as little code as possible. That's generally a good thing, but languages are too different that you could rule them all with one query, you get too many edge cases you have to wrestle with.

When he abandoned it I forked the project ([HiPhish/nvim-ts-rainbow2](https://github.com/HiPhish/nvim-ts-rainbow2)) in March this year. My fork split queries and strategies out from the core. This means we have to write a completely new query for each language, but it also means fewer edge cases.

And then I forked that fork again in July of this year because of changes in nvim-treesitter. And just today I have merged a massive refactoring. It's beautiful, finally HTML tags can have proper highlighting. Highlighting delimiters is not hard in itself, but there are weird edge cases when it comes to nested languages.

Please don't misunderstand me, I'm not whining. I have wanted a plugin like this even back when I was using Vim before Neovim even existed, I just never imagined I would be the one maintaining it. I find the journey fascinating.

PR #54, opened 2023-11-11
Add more LuaLS types for better LSP info

Danielkonge:
I added a lot of type information, so we can get proper LSP information when setting up the plugin:

Screenshot 2023-11-11 at 17 42 35

Screenshot 2023-11-11 at 17 43 05

Screenshot 2023-11-11 at 17 43 24

Screenshot 2023-11-11 at 17 43 46

I also made a few other small edits that might be worth discussing. Here is a list of things to think about before this is ready to be committed:

- [x] Right now I allow random strings in place of a known language, strategy, highlight color or query. This is because a user of the plugin could define their own, but this also means that the LSP won't really help out with spelling mistakes, since e.g. `vim` misspelled as `vin` is still a string. We could change this or see if there is some good workaround. (Spelling mistakes will be caught by `:checkhealth`, so we should be good with the current code.)
- [x] Since we don't want `setup` in `rainbow-delimiters.lua`, I have added an `__index` meta method for now. This is a temporary solution, that I think should make lazy work as expected, but I am not sure if that is a good solution. I assume you would still like to avoid putting `setup` in `rainbow-delimiters.lua`? (Think about this later.)
- [x] I think the standard for Neovim is to use `integer` instead of `number` for stuff like `bufnr` (or anything else that you know is an integer). Do you want to change `number` to `integer` or leave it as it is now? (Yes, use `integer`.)
- [x] I added `'comment'` to the blacklist, since that seems like a good idea based on #53. I can remove it again, if you prefer? (Remove it for now and instead investigate why `comment` causes problems.)

Other notes:
- If you ever do a rewrite, note that it is easier to give types for strings that don't need to be inside `['<name>']` in a Lua table (like `['local']` or `['']`).
- I couldn't figure out a way for the plugin to capture the fact that the type of `vim.g.rainbow_delimiters` should be of type `rainbow_delimiters.config`. So for now a user of the plugin will have to add `---@type rainbow_delimiters.config` before `vim.g.rainbow_delimiters` to get the LSP to work as expected. I have added this to the docs for now, but we can remove it if we figure out how to avoid this step. If you use the setup function the type is already correct now without the user adding anything. 
- I haven't figured out a way to remove used strings from the arrays with strings (like highlight or blacklist). So for those the LSP will continue to suggest a string that has already been used right now, but this should be easy enough to work with still.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -50,7 +50,7 @@ Configuration is done by setting entries in the Vim script dictionary
            \ 'RainbowDelimiterGreen',
            \ 'RainbowDelimiterViolet',
            \ 'RainbowDelimiterCyan',
-       \ ], 
+       \ ],
    \ }
 
 The equivalent code in Lua:
@@ -60,6 +60,7 @@ The equivalent code in Lua:
    -- This module contains a number of default definitions
    local rainbow_delimiters = require 'rainbow-delimiters'
 
+   ---@type rainbow_delimiters.config
    vim.g.rainbow_delimiters = {
        strategy = {
            [''] = rainbow_delimiters.strategy['global'],
@@ -87,7 +88,7 @@ same parameters as `g:rainbow-delimiters`.
 .. code:: lua
 
    require('rainbow-delimiters.setup').setup {
-       strategy = { 
+       strategy = {
            -- ...
        },
        query = {
--- a/doc/rainbow-delimiters.txt
+++ b/doc/rainbow-delimiters.txt
@@ -108,6 +108,7 @@ Alternatively, the same configuration in Lua:
     -- This module contains a number of default definitions
     local rainbow_delimiters = require 'rainbow-delimiters'
 
+    ---@type rainbow_delimiters.config
     vim.g.rainbow_delimiters = {
         strategy = {
             [''] = rainbow_delimiters.strategy['global'],
@@ -423,9 +424,9 @@ object-oriented terminology we would say that a strategy table must implement
 the strategy protocol.
 >
     strategy = {
-        on_attach = function(bufnr: number, settings: table),
-        on_detach = function(bufnr: string),
-        on_reset  = function(bufnr: number, settings: table),
+        on_attach = function(bufnr: integer, settings: table),
+        on_detach = function(bufnr: integer),
+        on_reset  = function(bufnr: integer, settings: table),
     }
 <
 
--- a/lua/rainbow-delimiters.lua
+++ b/lua/rainbow-delimiters.lua
@@ -18,7 +18,7 @@
 local lib = require 'rainbow-delimiters.lib'
 
 ---Disable rainbow delimiters for a given buffer.
----@param bufnr number  Buffer number, zero for current buffer.
+---@param bufnr integer  Buffer number, zero for current buffer.
 local function disable(bufnr)
 	if not bufnr or bufnr == 0 then bufnr = vim.api.nvim_get_current_buf() end
 	lib.detach(bufnr)
@@ -26,13 +26,15 @@ local function disable(bufnr)
 end
 
 ---Enable rainbow delimiters for a given buffer.
----@param bufnr number  Buffer number, zero for current buffer.
+---@param bufnr integer  Buffer number, zero for current buffer.
 local function enable(bufnr)
 	if not bufnr or bufnr == 0 then bufnr = vim.api.nvim_get_current_buf() end
 	lib.buffers[bufnr] = nil
 	lib.attach(bufnr)
 end
 
+---Toggle rainbow delimiters for a given buffer.
+---@param bufnr integer  Buffer number, zero for current buffer.
 local function toggle(bufnr)
 	if not bufnr or bufnr == 0 then bufnr = vim.api.nvim_get_current_buf() end
 	if lib.buffers[bufnr] then
@@ -47,8 +49,11 @@ local M = {
 	hlgroup_at = lib.hlgroup_at,
 	---Available default highlight strategies
 	strategy = {
+		---Global highlighting strategy
 		['global'] = require 'rainbow-delimiters.strategy.global',
+		---Local highlighting strategy
 		['local']  = require 'rainbow-delimiters.strategy.local',
+		---Empty highlighting strategy for testing
 		['noop']   = require 'rainbow-delimiters.strategy.no-op',
 	},
 	enable  = enable,
--- a/lua/rainbow-delimiters/lib.lua
+++ b/lua/rainbow-delimiters/lib.lua
@@ -55,6 +55,7 @@ M.nsids = setmetatable({}, {
 ---is a table of information about that buffer (e.g. language, strategy,
 ---query).  This also makes sure we keep track of all parsers in active use to
 ---prevent them from being garbage-collected.
+---@type table<integer, rainbow_delimiters.buffer_settings | false>
 M.buffers = {}
 
 
@@ -78,11 +79,10 @@ function M.get_query(lang)
 end
 
 ---Apply highlighting to a single node.
----@param bufnr   number  Buffer which contains the node
+---@param bufnr   integer  Buffer which contains the node
 ---@param lang    string  Language of the node (to group HL into namespaces)
 ---@param node    table   Node to highlight
 ---@param hlgroup string  Name of the highlight group to  apply.
----@return nil
 function M.highlight(bufnr, lang, node, hlgroup)
 	-- range of the capture, zero-indexed
 	local startRow, startCol, endRow, endCol = node:range()
@@ -103,7 +103,7 @@ end
 
 
 ---Get the appropriate highlight group for the given level of nesting.
----@param i number  One-based index into the highlight groups
+---@param i integer  One-based index into the highlight groups
 ---@return string hlgroup  Name of the highlight groups
 function M.hlgroup_at(i)
 	local hlgroups = config.highlight
@@ -113,8 +113,10 @@ end
 
 ---Clears the reserved Rainbow namespace.
 ---
----@param bufnr number  Number of the buffer for which to clear the namespace
----@return nil
+---@param bufnr integer  Number of the buffer for which to clear the namespace
+---@param lang string
+---@param line_start integer?
+---@param line_end integer?
 function M.clear_namespace(bufnr, lang, line_start, line_end)
 	local nsid = M.nsids[lang]
 	if vim.api.nvim_buf_is_valid(bufnr) then
@@ -123,6 +125,7 @@ function M.clear_namespace(bufnr, lang, line_start, line_end)
 end
 
 ---Start rainbow highlighting for the given buffer
+---@param bufnr integer
 function M.attach(bufnr)
 	-- Rainbow delimiters was explicitly disabled for this buffer
 	if M.buffers[bufnr] == false then return end
@@ -175,10 +178,12 @@ function M.attach(bufnr)
 	if not strategy or strategy == vim.NIL then return end
 
 	parser:register_cbs {
+		---@param bnr integer
 		on_detach = function(bnr)
 			if not M.buffers[bnr] then return end
 			M.detach(bufnr)
 		end,
+		---@param child LanguageTree
 		on_child_removed = function(child)
 			M.clear_namespace(bufnr, child:lang())
 		end,
@@ -201,6 +206,7 @@ function M.attach(bufnr)
 end
 
 ---Start rainbow highlighting for the given buffer
+---@param bufnr integer
 function M.detach(bufnr)
 	log.trace('Detaching from buffer %d.', bufnr)
 	if not M.buffers[bufnr] then
--- a/lua/rainbow-delimiters/stack.lua
+++ b/lua/rainbow-delimiters/stack.lua
@@ -17,18 +17,22 @@
 ---Helper library for stack-like tables.
 local M = {}
 
----@class Stack
----@field public  size    fun(self: Stack): number
+---@class (exact) Stack
+---@field public  size    fun(self: Stack): integer
 ---@field public  peek    fun(self: Stack): any
 ---@field public  push    fun(self: Stack, item: any): Stack
 ---@field public  pop     fun(self: Stack): any
----@field public  iter    fun(self: Stack): (fun(i: number, item: any): number, any), Stack, number
----@field private content any[]
+---@field public  iter    fun(self: Stack): ((fun(i: integer, item: any): integer?, any), Stack, integer)
+---@field package content any[]
 
 ---The stack metatable.
 local mt = {}
 
 ---The actual iterator implementation, hidden behind the iter-method.
+---@param stack Stack
+---@param i integer
+---@return integer?
+---@return any
 local function iter_stack(stack, i)
 	if i <= 1 then return end
 	return i - 1, stack.content[i - 1]
@@ -39,7 +43,7 @@ end
 local function stack_tostring(stack)
 	local items = {}
 	for _, item in ipairs(stack.content) do
-		items[#items+1] = tostring(item)
+		items[#items + 1] = tostring(item)
 	end
 	return string.format('[%s]', table.concat(items, ', '))
 end
@@ -48,7 +52,8 @@ end
 ---[ Methods ]-----------------------------------------------------------------
 
 ---Returns the current number of items in the stack.
----@return number size  Current size of the stack
+---@param self Stack
+---@return integer size  Current size of the stack
 local function size(self)
 	return #self.content
 end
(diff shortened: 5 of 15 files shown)


HiPhish:
I am all in favour of more type hints, but I think you went a bit too far in typing out everything.

> Right now I allow random strings in place of a known language, strategy, highlight color or query. This is because a user of the plugin could define their own, but this also means that the LSP won't really help out with spelling mistakes, since e.g. `vim` misspelled as `vin` is still a string. We could change this or see if there is some good workaround.

I don't think we need to type-check arbitrary strings. This is already handled by `:checkhealth` at runtime: for each language listed check if there is a corresponding grammar installed and report an error if there isn't.

```
==============================================================================
rainbow-delimiters: require("rainbow-delimiters.health").check()

Custom strategies ~
- OK Valid custom strategy for 'commonlisp'.
- ERROR No parser installed for 'asdfwerwer'
```

How would your suggestion handle language names we don't have registered in the plugin?

> Since we don't want `setup` in `rainbow-delimiters.lua`, I have added an `__index` meta method for now. This is a temporary solution, that I think should make lazy work as expected, but I am not sure if that is a good solution. I assume you would still like to avoid putting setup in `rainbow-delimiters.lua`?

This is a whole can of worms I have been kicking down the road. If the user calls a `setup` function (or any function) and that function in return has `require` it kicks off a whole chain of `require` calls, each of which has to traverse the file system to find the module. That is one of the reasons why `setup` functions are bad, and people have been working their way around that with lazy-loading package managers, people are trying to solve a problem they should not be having in the first place. This is why it is better to have a `g:` variable instead, defining variables is cheap in comparison.

I will have to look into this issue deeper because my code still does `require` calls that I should avoid.

> I think the standard for Neovim is to use `integer` instead of `number` for stuff like `bufnr` (or anything else that you know is an integer). Do you want to change `number` to `integer` or leave it as it is now?

Can the language server handle `integer`? I know Lua 5.1 only has a `number` type, integers were only added in Lua 5.3. If the server can handle `integer` then that would be a better annotation, the more accurate information is, the better.

> I added 'comment' to the blacklist, since that seems like a good idea based on https://github.com/HiPhish/rainbow-delimiters.nvim/issues/53. I can remove it again, if you prefer?

I would rather know why `comment` needs to be blacklisted in the first place, since we don't have any query for it. The problem with populating the blacklist by default is what happens if the user has his own blacklist. Does the user's blacklist override the default one? Do they get merged? If they get merged, what if the user for whatever reason actually does want `comment` to work?

> If you ever do a rewrite, note that it is easier to give types for strings that don't need to be inside `['<name>']` in a Lua table (like `['local']` or `['']`).

Please, no more rewrites :( Yeah, `local` was maybe not the best choice. Instead of the empty string I originally wanted the index `1`, which looked really cool in Lua: `{foo, x = 'bar', y = 'baz'}`. This is similar to `*args` and `**kwargs` in Python. Unfortunately tables which have both an array- and a dict part do not convert well between Lua and Vim script.

Danielkonge:
> I am all in favour of more type hints, but I think you went a bit too far in typing out everything.

This is really more of a draft where I played around with the types a bit. I plan to clean up everything a bit soon. The main focus is what I did in `rainbow-delimiters.types.lua`. (I changed the status to draft in Github for now.)

> I don't think we need to type-check arbitrary strings. This is already handled by `:checkhealth` at runtime: for each language listed check if there is a corresponding grammar installed and report an error if there isn't.
> 
> ```
> ==============================================================================
> rainbow-delimiters: require("rainbow-delimiters.health").check()
> 
> Custom strategies ~
> - OK Valid custom strategy for 'commonlisp'.
> - ERROR No parser installed for 'asdfwerwer'
> ```
> 
> How would your suggestion handle language names we don't have registered in the plugin?

I am not really including the types for the strings in e.g. `whitelist` or `blacklist` to type check them, it is more to get the completions to help writing your lists. I think the current system works fine, so maybe the current solution is good enough. You will get completions while writing e.g. `commonlisp`, and if you don't use the completion and accidentally write `comonlisp` or similar, then `:checkhealth` will catch it. 

Do you not think it is worth it to include a list of the language strings to give you completions?
 
> This is a whole can of worms I have been kicking down the road. If the user calls a `setup` function (or any function) and that function in return has `require` it kicks off a whole chain of `require` calls, each of which has to traverse the file system to find the module. That is one of the reasons why `setup` functions are bad, and people have been working their way around that with lazy-loading package managers, people are trying to solve a problem they should not be having in the first place. This is why it is better to have a `g:` variable instead, defining variables is cheap in comparison.
> 
> I will have to look into this issue deeper because my code still does `require` calls that I should avoid.

Ok, for now I will remove this code from this pull request then. It sounds like we should think about that more.

> Can the language server handle `integer`? I know Lua 5.1 only has a `number` type, integers were only added in Lua 5.3. If the server can handle `integer` then that would be a better annotation, the more accurate information is, the better.

[LuaLS can handle `integer` no matter which version of Lua you use](https://luals.github.io/wiki/annotations/#documenting-types) (i.e., for 5.1-5.4 + LuaJIT)

> I would rather know why `comment` needs to be blacklisted in the first place, since we don't have any query for it. The problem with populating the blacklist by default is what happens if the user has his own blacklist. Does the user's blacklist override the default one? Do they get merged? If they get merged, what if the user for whatever reason actually does want `comment` to work?

I will remove `comment` from the blacklist for now and try to see if I can figure out why it causes problems instead. 

> Please, no more rewrites :( Yeah, `local` was maybe not the best choice. Instead of the empty string I originally wanted the index `1`, which looked really cool in Lua: `{foo, x = 'bar', y = 'baz'}`. This is similar to `*args` and `**kwargs` in Python. Unfortunately tables which have both an array- and a dict part do not convert well between Lua and Vim script.

I don't plan on doing work towards a rewrite either. :/ It was more of a general comment. 


I will add some of the updates later tonight.

Danielkonge:
I have cleaned up this draft a bit now (removed superfluous type annotations and removed the `__index` code and `comment` blacklist mentioned above). 

Please let me know if there is anything in the types you would want me to simplify or remove. I think the current setup in `rainbow-delimiters.types.lua` gives me good suggestions/completions when setting up the plugin, but I can easily change it.

README.rst
```diff
@@ -80,14 +81,17 @@ The equivalent code in Lua:
        },
    }
 
+Note: The line with `---@type rainbow_delimiters.config` is optional, but it
+will give you better completions with LuaLS.
+
```
HiPhish, line 86: I don't think we need to mention this. It is obvious that comments are always optional, and anyone using this plugin is expected to know at least that much Lua.

doc/rainbow-delimiters.txt
```diff
@@ -130,6 +131,10 @@ Alternatively, the same configuration in Lua:
     }
 <
 
+Note: The line with `---@type rainbow_delimiters.config` is optional, but it
+will give you better completions with LuaLS.
+
+
```
HiPhish, line 137: Same as my comment in the README

doc/rainbow-delimiters.txt
```diff
@@ -157,7 +162,7 @@ available as part of a Lua module.
         blacklist = {'c', 'cpp'},
     }
 <
-The keys are exactly the same as for |g:rainbow_parentheses|.  If fact, this
+The keys are exactly the same as for |vim.g.rainbow_delimiters|.  In fact, this
```
HiPhish, line 165: Thanks for catching this mistake, but please keep global Vim variables as `g:` variables because that is what the manual does.

```suggestion
The keys are exactly the same as for |g:rainbow_delimiters|.  In fact, this
```

lua/rainbow-delimiters.lua
```diff
@@ -56,6 +59,5 @@ local M = {
 	toggle  = toggle,
 }
 
-
```
HiPhish, line 59: I like that extra space before the last statement group. It gives me visual separation between the body of the script and the closing statements.

lua/rainbow-delimiters/lib.lua
```diff
@@ -55,6 +55,7 @@ M.nsids = setmetatable({}, {
 ---is a table of information about that buffer (e.g. language, strategy,
 ---query).  This also makes sure we keep track of all parsers in active use to
 ---prevent them from being garbage-collected.
+---@type table<integer, rainbow_delimiters.buffer_settings | false | nil>
```
HiPhish, line 58: Do we need `nil` here?

HiPhish (changes requested):

HiPhish:
I have reviewed all changes except for `lua/rainbow-delimiters.types.lua` and `lua/rainbow-delimiters/stack.lua`, I will take a bit more time for those. I noticed that you have also made some corrections to the code and the manual. Ideally those should be separate PRs so we can merge then sooner, but it's OK for now.

doc/rainbow-delimiters.txt
Danielkonge: Ok, I will change it back. Just as a note: The reason I changed it here was to distinguish between the vim code and Lua code above it, where the Lua code uses `vim.g.rainbow_delimiters`, but is should be pretty obvious what is meant here anyways.

README.rst
Danielkonge: Okay, I have removed it now. I am often unsure what exactly one expects the users to know.

lua/rainbow-delimiters.lua
Danielkonge: I have added the space again. I think I accidentally removed it when removing the previously mentioned `__index` code.

lua/rainbow-delimiters/lib.lua
Danielkonge: We don't need it, so whether we should keep it depends on how you like to think about setting something to `nil` (as "unsetting" it or as literally setting it to `nil`).

We do have `M.buffers[bufnr] = nil` later in the code, which is why I added this, so we are descriptive about what we might set values as. But LuaLS is fine with setting it to `nil` without having it in the type (I assume because `nil` values are ignored in tables anyways, so it really is like unsetting it).

Danielkonge:
> I have reviewed all changes except for `lua/rainbow-delimiters.types.lua` and `lua/rainbow-delimiters/stack.lua`, I will take a bit more time for those. I noticed that you have also made some corrections to the code and the manual. Ideally those should be separate PRs so we can merge then sooner, but it's OK for now.

I thought it was mostly minor things that might as well be included here, but if you prefer I can quickly move the small cleanup stuff to another pull request.

HiPhish:
How does the file `rainbow-delimiters.types.lua` work? Is it used automatically by the language server or do I need to specify some project settings (e.g. in `.luarc.json`)?

Danielkonge:
> How does the file `rainbow-delimiters.types.lua` work? Is it used automatically by the language server or do I need to specify some project settings (e.g. in `.luarc.json`)?

neodev should ensure that everything is set up correctly for all plugins by default. 

Without neodev I think you need to set your runtime path correctly in a `.luarc.json` or in your Lua LSP setup.

lua/rainbow-delimiters/lib.lua
Danielkonge: Looking at this again, I do think it makes sense to remove the `nil` from the type, since I mostly do think of setting a table value to nil as unsetting it.

Danielkonge:
Sorry the sync on this branch got messed up. I will fix it soon.

lua/rainbow-delimiters/stack.lua
```diff
@@ -98,12 +103,12 @@ end
...
-		iter = iter,
-		push = push,   ---@type fun(self: Stack, item: any): Stack
-		pop  = pop,    ---@type fun(self: Stack): any
-		peek = peek,   ---@type fun(self: Stack): any
+		content = {}, ---@type any[]
+		size    = size, ---@type fun(self: Stack): integer
+		iter    = iter, ---@type fun(self: Stack): ((fun(i: integer, item: any): integer?, any), Stack, integer)
+		push    = push, ---@type fun(self: Stack, item: any): Stack
+		pop     = pop, ---@type fun(self: Stack): any
+		peek    = peek, ---@type fun(self: Stack): any
```
HiPhish, line 111: Do we still need these type annotations? I think we can remove them, the functions and the fields are all type annotated already.

```suggestion
		content = {},
		size    = size,
		iter    = iter,
		push    = push,
		pop     = pop,
		peek    = peek,
```

lua/rainbow-delimiters/stack.lua
```diff
@@ -18,17 +18,21 @@
 local M = {}
 
 ---@class Stack
```
HiPhish, line 20: We can also mark the class as "exact", meaning that there can be no additional fields. Do you agree?

```suggestion
---@class (exact) Stack
```

HiPhish:
I'm done with `stack.lua`, this just leaves the type definitions file.

lua/rainbow-delimiters/stack.lua
Danielkonge: I think you are right, we don't need these. I only added the last one since the others were already there.

lua/rainbow-delimiters/stack.lua
Danielkonge: Yes, we can do that. In principle we could probably do that with most classes in the type definitions, but for things that might change I think it makes sense to not include exact. Here Stack is stable though, so it makes a lot of sense to include it.

HiPhish:
Two notes:

- I removed the whitespace from the description because it shows up in the hover window
- All configuration fields are optional; this plugins works even without any configuration

Danielkonge:
> Two notes:
> 
> * I removed the whitespace from the description because it shows up in the hover window
> * All configuration fields are optional; this plugins works even without any configuration

In that case I guess `['']` should also be optional in the strategies and queries. I have updated this now.

HiPhish:
> Also, should we maybe add a note in the docs and/or README about making sure to set up the LSP correctly (e.g. via Neodev)?

I don't think so. People either already know how to set up the Lua language server correctly, in which case that's just noise, or they don't, in which case that's just useless and confusing bloat.

I feel the same about plugin installation instructions that list standard code snippets for five different plugin managers, when it should be just one sentence that says "install it like any other plugin". Unless the plugin has special requirements (compiled binary, dependencies, extra configuration steps) of course. Maybe my brain is wired differently, but when I keep seeing the same text over and over again by brain starts to block out that boilerplate, and that's when mistakes are made because there was one time where it was actually useful, but I did not see it.

HiPhish:
I have removed the class definition for the plugin library. The language server can derive the type of all fields from their values, and we never pass the table to any other function. Is there something that would work better if we had a separate class for the table?

HiPhish:
I don't really like that we have to list all supported languages three times. I tried this annotations:

```
---@field query table<rainbow_delimiters.language, string>?
```

In my mind this should work, the key type is `rainbow_delimiters.language`, so the language server should suggest on the valid values, but it doesn't suggest any of them. Do you have an idea how we could reuse the `rainbow_delimiters.language` type? Otherwise will will have to put up with the huge list.

HiPhish:
I have removed the type for highlight groups. People will either use the default list, in which case they don't need auto-completion, or they will use custom highlight groups from some theme, in which case the auto-completion is actually counter-productive.

HiPhish (approved):

HiPhish:
I have nothing else to add, so if you are done as well let's squash and merge everything.

> But this seems more like a LuaLS bug than an actual problem with the types.

Yes, I noticed it as well, but it should be handled by the language server.

Danielkonge:
> I don't really like that we have to list all supported languages three times. I tried this annotations:
> 
> ```
> ---@field query table<rainbow_delimiters.language, string>?
> ```
> 
> In my mind this should work, the key type is `rainbow_delimiters.language`, so the language server should suggest on the valid values, but it doesn't suggest any of them. Do you have an idea how we could reuse the `rainbow_delimiters.language` type? Otherwise will will have to put up with the huge list.

It is hard to avoid multiple "duplicate" types like the current setup, because LuaLS doesn't really allow us to construct the types programatically. I don't know any good way than the current definitions that gives us proper completions.

Something like 
```lua
---@class rainbow_delimiters.config.strategies
---@field [''] (rainbow_delimiters.strategy | fun(): rainbow_delimiters.strategy?)?
---@field [rainbow_delimiters.language] (rainbow_delimiters.strategy | fun(): rainbow_delimiters.strategy?)?
```
should give the correct type checks, but not the correct completions, which I think is a big part of the help from the LSP, and I think it is probably the same for what you wrote; you will get the type checks but not completions.

Danielkonge:
> I have nothing else to add, so if you are done as well let's squash and merge everything.

I have squashed it now.

Danielkonge:
One note before you commit it:

HiPhish:
Either way is fine by me. I like `.global` better, but since we cannot have `.local` let's be consistent.

Danielkonge:
> Either way is fine by me. I like `.global` better, but since we cannot have `.local` let's be consistent.

Okay, I have made it `['global']` now to be consistent (since we can't have `.local` as you said), and I have squashed the commits again.

Danielkonge:
I added a note on types to `CONTRIBUTING.rst` in a separate commit. I can squash this commit too if you prefer. Also, if you commit the luadoc pull request first, I can add its types in this commit.

Danielkonge:
I added the luadoc types mentioned above now.

HiPhish:
I have merged it now, thank you.

PR #55, opened 2023-11-12
feat(nim): added queries and tests

aMOPel:


HiPhish:
How and from there did you install the Nim parser? When I execute `TSInstall nim` I get `Parser not available for language "nim"`.

aMOPel:
The pr to add nim to nvim-treesitter was only merged 2 weeks ago. Maybe you need to update you plugin?

queries/nim/rainbow-delimiters.scm
```diff
@@ -0,0 +1,19 @@
+(array_construction "[" @delimiter "]" @delimiter @sentinel) @container
+(tuple_construction "(" @delimiter ")" @delimiter @sentinel) @container
+(tuple_deconstruct_declaration "(" @delimiter ")" @delimiter @sentinel) @container
+(curly_construction "{" @delimiter ":"? @intermediate "}" @delimiter @sentinel) @container
```
Danielkonge, line 4:
```suggestion
(curly_construction "{" @delimiter ":"? @delimiter "}" @delimiter @sentinel) @container
```

Danielkonge (commented):
You should be able to use `@delimiter` for everything being highlighted; `@intermediate` will not work.

HiPhish:
I have made a couple of changes:

- Formatted the patterns to be more like other queries
- Renamed the test file to be more like the other test files
- Fixed the `cast` pattern by merging both patterns into one

We are no longer bound to the limitation of just one opening and one closing node per delimiter. Both the `[`, `]` and the `(`, `)` can be part of the same container on the same level.

```
(cast
  type: "[" @delimiter
  type: "]" @delimiter
  value: "(" @delimiter
  value: ")" @delimiter @sentinel) @container
```

If you agree with my changes please squash the commits into one, then I can merge your PR.

aMOPel:
Thank you :)

All done.

HiPhish:
Merged, thank you very much.

PR #56, opened 2023-11-12
Very minor cleanup

Danielkonge:
I noticed that `CONTRIBUTING.rst` still said `@intermediate` instead of `@delimiter`.

Also, I remember discussing updating the version number after the big update. I added it here mainly as a reminder, but I can just remove it again if you prefer.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -15,10 +15,10 @@ manual first to understand how to write queries.  Your query should meet the
 following criteria:
 
 - Named `rainbow-delimiters`
-- Few if any `@intermediate` capture groups; we do not want the default query
-  to be too vibrant
+- Few `@delimiter` capture groups per `@container`; we do not want the default
+  query to be too vibrant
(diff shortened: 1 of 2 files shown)


HiPhish:
Merged, thank you. I have tagged the new version now.

Danielkonge:
By the way, I think we should add some new screenshots with this update. Especially Lua with the blocks queries and html as a whole has much nicer highlighting now. I can take some and make a new pull request if you would want that.

HiPhish:
I would rather take them myself to ensure the font and theme are the same in all screenshots. It's not much work. Unless you really feel the need to replicate my setup yourself that is :)

Danielkonge:
> I would rather take them myself to ensure the font and theme are the same in all screenshots. It's not much work. Unless you really feel the need to replicate my setup yourself that is :)

I will just leave it to you then. :)

I just thinks some of the highlighting looks much better now, so it is worth highlighting those cases with new pictures.

PR #57, opened 2023-11-14
Add comment on `html-tag-names` queries

Danielkonge:
I saw a [request to be able to keep the old style html highlighting](https://www.reddit.com/r/neovim/comments/17tiq3a/big_change_to_rainbowdelimitersnvim/) on reddit, so I thought I would add it. Please let me know if you have a better name for it.

### DIFF
--- a/queries/html/rainbow-delimiters.scm
+++ b/queries/html/rainbow-delimiters.scm
@@ -1,6 +1,14 @@
 ;;; A pair of delimiter tags with any content in-between.
 ;;; Last tag should be a sentinel.
 
+;;; If instead you want rainbow-delimiters to only highlight
+;;; the tag names without any of "<", "</", ">" or "/>", then
+;;; you can make your own query file, e.g.,
+;;;   'rainbow-tag-names'
+;;; and use the following with
+;;;   x @delimiter
+;;; deleted for x equal to any of "<", "</", ">" or "/>".
+
 (element
   (start_tag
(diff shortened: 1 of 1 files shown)


HiPhish:
Do we really want such a query included by default? That's a very niche use-case and I don't think anyone else will want it. Plus, if we include it I would like the same for React as well.

Danielkonge:
> Do we really want such a query included by default? That's a very niche use-case and I don't think anyone else will want it. Plus, if we include it I would like the same for React as well.

I guess it will add extra stuff to maintain that might not be used by a lot of people. Do you think it would make sense if I instead add a brief comment about how to make this in the html queries file?

HiPhish:
> I guess it will add extra stuff to maintain that might not be used by a lot of people.

Right. Tree-sitter queries don't have any mechanism for abstraction, so there will be a lot of copy-pasting that needs to be kept in sync manually. If certain patterns are commonly used it makes sense to do it, but such a niche case is just an extra maintenance burden. This is precisely the reason why I wanted to offer an escape hatch to people from day one. Whatever weird or silly query or strategy you have, you can add it yourself to your own config.

> Do you think it would make sense if I instead add a brief comment about how to make this in the html queries file?

That sounds like a good idea to me.

DeadlySquad13:
Thanks for attention to my comment!
Don't really understand why everyone don't use it like I do but well :D...

PR #62, opened 2023-11-18
TS/JS switch statement missing

bomgar:
Hi,

I notices the braces around the switch statement body are not colorized.  I tried to fix it but for some reason it only works in typescript and not in javascript. Any ideas?

### DIFF
--- a/queries/javascript/rainbow-delimiters.scm
+++ b/queries/javascript/rainbow-delimiters.scm
@@ -18,6 +18,10 @@
   "{" @delimiter
   "}" @delimiter @sentinel) @container
 
+(switch_body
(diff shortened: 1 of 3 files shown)


HiPhish:
Merged, thank you.

Danielkonge:
> Hi,
> 
> I notices the braces around the switch statement body are not colorized. I tried to fix it but for some reason it only works in typescript and not in javascript. Any ideas?

Do you still have this problem with it not working in javascript? For me this seems to work in both javascript and typescript. Are your parsers updated?

bomgar:
I don't have the problem anymore. I have no idea what caused this but it's gone.

PR #63, opened 2023-11-19
Kotlin

bomgar:
add kotlin support

### DIFF
--- a/queries/kotlin/rainbow-delimiters.scm
+++ b/queries/kotlin/rainbow-delimiters.scm
@@ -1,0 +1,48 @@
+(class_body
+  "{" @delimiter
+  "}" @delimiter @sentinel) @container
+
+(function_body
+  "{" @delimiter
+  "}" @delimiter @sentinel) @container
+
+(control_structure_body
+  "{" @delimiter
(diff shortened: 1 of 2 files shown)


HiPhish:
Thank you. I noticed that the curly brace around the `when` body are not included. However, they are in the hierarchy above the `when_subject`, so the `when_subject` would be highlighted one level below the body, even though optically they are on the same level and the subject comes before the body. That's just weird.

How about this pattern instead?

```
(when_expression
  subject: (when_subject
    "(" @delimiter
    ")" @delimiter)
  "{" @delimiter
  "}" @delimiter @sentinel) @container
```

This would highlight both sets of delimiters at the same level. Does this make sense in Kotlin, or are there other cases where this would look wrong?

bomgar:
I think that makes sense. You want me to open another PR?

bomgar:
I totally forgot about that one:
```kotlin
when {
    1 == 1 -> print("1")
    else -> print("not")
}

```

HiPhish:
It's OK, I have done it myself. I just needed confirmation from someone who knows Kotlin better.

PR #64, opened 2023-11-19
[fix] kotlin: second when syntax missing

bomgar:
Sorry totally forgot the second when syntax.


bomgar:
I guess both expressions are required

HiPhish:
I have already fixed it by making the `when_subject` optional. Your solution assigns the wrong level to the first `when` (yellow in my screenshot instead of red).

Screenshot_20231119_163940

bomgar:
Ah. Thank you.

PR #65, opened 2023-11-19
Add luadoc queries

Danielkonge:
The luadoc parser has a few problems with things it captures weirdly (or doesn't really capture), but we can get good highlighting for most type annotations in Lua with this. 

Note: If you commit this before the update with types, then I can add `luadoc` there. Otherwise, if you commit the types update first, I can rebase this pull request and add the types here.

### DIFF
--- a/queries/luadoc/rainbow-delimiters.scm
+++ b/queries/luadoc/rainbow-delimiters.scm
@@ -1,0 +1,39 @@
+(function_type
+  "(" @delimiter
+  ")" @delimiter @sentinel
+) @container
+
+(parenthesized_type
+  "(" @delimiter
+  ")" @delimiter @sentinel
+) @container
+
+;;; I wanted to use something like
+; (union_type
+;   "|" @delimiter @sentinel
+; ) @container
+;;; too, but it doesn't fully work with the current parser
+
+(array_type
+  "[" @delimiter
+  "]" @delimiter @sentinel
+) @container
+
+(table_type
+  "<" @delimiter
+  ">" @delimiter @sentinel
+) @container
+
+(table_literal_type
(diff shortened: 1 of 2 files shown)


HiPhish:
Merged. What does this pattern mean?

```
(_
  "[" @delimiter
  .
  field: (_)
  .
  "]" @delimiter @sentinel
) @container
```

In this text does `[integer]: integer` mean a table where there is any number of integer keys that map to integer values?

```
---@type { key1: { key2: { [string]: table<number, number | integer> }, [integer]: integer } }
```

Danielkonge:
> Merged. What does this pattern mean?
> 
> ```
> (_
>   "[" @delimiter
>   .
>   field: (_)
>   .
>   "]" @delimiter @sentinel
> ) @container
> ```

The dots just indicate that the elements should show up as immediate neighbors, and the reason I use `_` twice is that there are multiple cases, where you write something like `[integer]` or `[<any_type>]` really. Instead of writing out the different cases, this should capture all of them. (I don't think the dots are actually needed, but I included them to be sure I catch only the things I wanted.)

> In this text does `[integer]: integer` mean a table where there is any number of integer keys that map to integer values?
> ```
> ---@type { key1: { key2: { [string]: table<number, number | integer> }, [integer]: integer } }
> ```

Yes, that is my understanding of the types, e.g. a Lua table like this `{ 1 = 1, 2 = 2, 3 = 3 }` can be described by the type annotation `{ [integer]: integer }`. Otherwise, without the square brackets, you are meant to write keys before the colon in the type annotations, so `{ integer: integer }` would mean you are using the key `integer` in the Lua table, e.g. `{ integer = 1 }`.

PR #67, opened 2023-11-25
Minor code formatting cleanup

Danielkonge:
I wanted to fix a `[integer]` typo to `[string]` in the new priority types, and thought I might as well cleanup the spaces in the other type definitions while doing that. 

Also a note: I think there have been quite a few additions since the 0.1.0 tag, so maybe it would make sense to tag 0.1.1 now. (This is unrelated to this PR.)


HiPhish:
Thanks.

> Also a note: I think there have been quite a few additions since the 0.1.0 tag, so maybe it would make sense to tag 0.1.1 now.

Yes, I want to tag a new release, but people constantly keep opening up new PRs ;)

PR #68, opened 2023-11-27
feat(query): add query for identifier

binhtran432k:
I see that it is very hard to keep track of the multiple levels of query identifiers. I created them because HTML, JSX, and TSX all have repeating elements and you use rainbow for it by default.

Demo:
image

### DIFF
--- a/queries/query/rainbow-delimiters.scm
+++ b/queries/query/rainbow-delimiters.scm
@@ -1,5 +1,6 @@
 (named_node
   "(" @delimiter
+  (identifier) @delimiter
   ")" @delimiter @sentinel) @container
 
 (grouping


Danielkonge:
This seems more like something that we could make into a `rainbow-blocks` query instead of adding it `rainbow-delimiters`, since this is not really the "standard" behavior of `rainbow-delimiters` in other languages. If we add this in a `rainbow-blocks` query, would it make sense to expand any of the three other queries too?

Also, note that this doesn't look that good in `:InspectTree<CR>`, which is also highlighted by the query parser. (Though that is subjective. It could look better if `InspectTree` actually had formatting more like the formatting used when writing queries.)

binhtran432k:
Closed because of duplication with #70

Danielkonge:
> Closed because of duplication with #70

I just want to clarify. I think the stuff in #70 will be moved to its own different queries (not under `query`) when neovim updates, but for now it uses the `query` queries because that is what `InspectTree` uses right now. If you would like something like this in the `query` queries, I think you should just change this pull request so that it puts the queries you currently have in `rainbow-delimiters.scm` into `rainbow-blocks.scm` and then leave `rainbow-delimiters.scm` as it was before. 

I am not sure if @HiPhish would want to add the extra queries, since it gives more stuff to maintain, but the `query` parser seems pretty small and stable, so this shouldn't bring too much extra work.

PR #69, opened 2023-11-27
Minor docs cleanup

Danielkonge:
I added `priority` in a few examples to make it look more like the other examples and then I removed a `@_tag_name` in an example (it should only be in the example after the one I removed). 

Also, I removed stuff from the `TODO.rst` that doesn't seem to be true anymore.


Danielkonge:
I think this is out of date, so I will close it now. Might as well make a new pull request, if we want to cleanup the docs.

PR #72, opened 2023-12-08
Add missing c query

Danielkonge:
I noticed that the `c` queries were missing `enumerator_list`, so I added it.


HiPhish:
Merged, thank you.

PR #74, opened 2023-12-15
Add more `c` queries

Danielkonge:
I found some more missing `c` queries. I don't actually know `c` that well (I just found these while going through the neovim codebase), so I can't write tests that won't have the LSP give some complaints. Feel free to edit this, if you are more used to `c` (or suggest changes and I will update it). 

Also, there is some problem with highlighting in macros (I made a comment about it in the test file). I am not sure what is wrong there and haven't had time to look into it in more details yet.

### DIFF
--- a/test/highlight/c/regular.c
+++ b/test/highlight/c/regular.c
@@ -1,6 +1,19 @@
 #include <stdio.h>
 
-#define MACRO 0
+
+#define PI 3.14
+/* These aren't highlight correctly. A problem with the parser? */
+#define TESTMACRO (-1)
+#define min(X,Y) ((X) < (Y) ? (X) : (Y))
+
+
+/* Declaration with parentheses, a function pointer */
+static void (*callback)(int);
+int c_init() { return 1; }
+
+/* Macro type specifier */
+#define Map int Foo
+static Map(char *c_str) {return 4;}
 
 typedef enum {
   E1,
@@ -17,6 +30,8 @@ struct Point2D {
 	int y;
 };
 
+/* Compound literal expression */
+struct Point2D v = (struct Point2D){ 0, 0 };
 
 /* A function definition */
 int add(int x, int y) {
--- a/queries/c/rainbow-delimiters.scm
+++ b/queries/c/rainbow-delimiters.scm
@@ -1,10 +1,10 @@
 (parameter_list
-   "(" @delimiter
-   ")" @delimiter @sentinel) @container
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
 
 (argument_list
-   "(" @delimiter
-   ")" @delimiter @sentinel) @container
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
 
 (parenthesized_expression
   "(" @delimiter
@@ -26,8 +26,8 @@
   "]" @delimiter @sentinel) @container
 
 (field_declaration_list
-   "{" @delimiter
-   "}" @delimiter @sentinel) @container
+  "{" @delimiter
+  "}" @delimiter @sentinel) @container
 
 (array_declarator
   "[" @delimiter
@@ -49,3 +49,19 @@
 (enumerator_list
   "{" @delimiter
   "}" @delimiter @sentinel) @container
+
+(macro_type_specifier
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
+
+(preproc_params
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
+
+(compound_literal_expression
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
+
+(parenthesized_declarator
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container


test/highlight/c/regular.c
```diff
@@ -2,6 +2,16 @@
 
 #define MACRO 0
 
+int c_init() { 1 }
```
HiPhish, line 5: I guess you want to return a fixed number?

```suggestion
int c_init() { return 1; }
```

test/highlight/c/regular.c
```diff
@@ -2,6 +2,16 @@
...
+// the following aren't highlight correctly,
+// is this a problem with the parser?
+#define TESTMACRO (-1)
+#define lang_init() c_init()
+#define min(X,Y) ((X) < (Y) ? (X) : (Y))
+
+static test(c_str) = lang_init();
+static Map(c_str) langs = MAP_INIT;
```
HiPhish, line 13: I don't understand what these two lines are supposed to mean. Where does `MAP_INIT` come from? Where does `c_str` come from? I guess `c_str` is the name of a parameter, if so you have to explicitly name its type in C. And I guess the assignment means you want to assign the "function" `lang_init` to `test`. That won't work First `lang_init` is a macro, it's just text that will be spliced in at some point, not a value; you have to splice the call into the function body inside curly braces. Second, if you want to actually assign a function to a variable, then that variable must be of a function pointer type, which is its own can of worms.

```suggestion
static test(char *c_str) {lang_init()};
static Map(char *c_str) langs {MAP_INIT;}
```

test/highlight/c/regular.c
```diff
@@ -2,6 +2,16 @@
 
 #define MACRO 0
 
+int c_init() { 1 }
+// the following aren't highlight correctly,
+// is this a problem with the parser?
+#define TESTMACRO (-1)
+#define lang_init() c_init()
+#define min(X,Y) ((X) < (Y) ? (X) : (Y))
```
HiPhish, line 10: Yeah, I have noticed that `parenthesized_expression` is not captured at all if it belongs to a macro, not even by a simple query like `(parenthesized_expression) @container`.

Danielkonge: I think it might have something to do with the macro being parsed via an injected language, but I am not sure. I think I remember noticing something like this some time ago while looking at injected languages. I can try to look into this later and hopefully find a solution, but I think it might be a parser issue.

test/highlight/c/regular.c
Danielkonge: That seems reasonable, yes.

test/highlight/c/regular.c
Danielkonge: This was mostly made by just copying stuff from the neovim source code and trying to make the parser give me the tree I wanted. Most of it probably doesn't fully make sense. I will try to clean it up a bit more tomorrow.

test/highlight/c/regular.c
HiPhish: The injected language is C itself. Try this minimal example:

```c
int main(int argc, char **argv) {
	return 0;
}
```

Now let's evaluate `vim.treesitter.get_parser():children()` and we get the empty table `{}`. Now let's add a preprocessor definition.

```c
#define PI 3.14

int main(int argc, char **argv) {
	return 0;
}
```

If we evaluate `vim.treesitter.get_parser():children()` again the language tree will have one child element of language `c`. C++ does something similar, except the injected language is `cpp`.

HiPhish:
I have gone ahead and fixed the syntax of the test file. It should now cover all the patterns you want to match. The `macro_type_specifier` still baffles me, I don't know what exactly it is supposed to represent. I can get it to show up if I use a macro both for the function name and return type, but why would anyone want to do such a thing?

Danielkonge:
> I have gone ahead and fixed the syntax of the test file. It should now cover all the patterns you want to match. 

Thank you, the fixes look good to me. I squashed the commits now, is there anything else you would want to update before committing to the main branch?

> The `macro_type_specifier` still baffles me, I don't know what exactly it is supposed to represent. I can get it to show up if I use a macro both for the function name and return type, but why would anyone want to do such a thing?

I am not sure what exactly it is supposed to represent, but it did show up in the neovim codebase when I looked through it (I am not sure if I can find it again easily though...), so it is at least something that is used.

test/highlight/c/regular.c
Danielkonge: Rust also uses a language injection of `rust` for macros, but highlighting works as expected for all cases I have seen there. Also, just to clarify, I did think it was injecting `c` code in the `#define` statement, I was just wandering if that is what is messing with the highlighting, but I haven't looked into this.

HiPhish:
Merged, thank you. BTW, sometimes highlighting inside a macro does work for me:
Screenshot_20231223_014804

Danielkonge:
> Merged, thank you. BTW, sometimes highlighting inside a macro does work for me: Screenshot_20231223_014804

That is weird, I don't think I have seen it working for me. Do you have any idea of what might have been different when it did work? In general, I still don't really know what causes the problems with the macros.

HiPhish:
It could be an issue with Neovim 0.10. The highlighting is correct on 0.9.4 for me.

Danielkonge:
> It could be an issue with Neovim 0.10. The highlighting is correct on 0.9.4 for me.

I can reproduce this. Highlighting also works in 0.9.4 for me, but not in 0.10. I can see that 0.9.5 was just released, I can test that later. Knowing that it comes from a recent update, we should be able to narrow it down a lot by looking at the differences between 0.9.4, 0.9.5 and 0.10.

PR #76, opened 2023-12-18
feat(elixir): added query for access_call

jcapblancq:
Hi there,

Before: 
image

After:
image

Thank you again for the plugin, very nice upgrade with html / xml lately! Let me know if you need to tweak anything.

### DIFF
--- a/queries/elixir/rainbow-delimiters.scm
+++ b/queries/elixir/rainbow-delimiters.scm
@@ -20,6 +20,10 @@
   "[" @delimiter
   "]" @delimiter @sentinel) @container
 
+(access_call
+  "[" @delimiter
+  "]" @delimiter @sentinel) @container
+
 (bitstring
   "<<" @delimiter
   ">>" @delimiter @sentinel) @container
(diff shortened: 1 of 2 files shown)


queries/elixir/rainbow-delimiters.scm
```diff
@@ -20,6 +20,10 @@
   "[" @delimiter
   "]" @delimiter @sentinel) @container
 
+(access_call
+  "[" @opening
+  "]" @closing @sentinel) @container
```
HiPhish, line 25: Capture names have change, we no longer differentiate between different kinds of delimiters. I am surprised the code worked for you at all.

```suggestion
  "[" @delimiter
  "]" @delimiter @sentinel) @container
```

test/highlight/elixir/regular.exs
```diff
@@ -22,6 +22,10 @@ defmodule Regular do
         IO.puts "The first value is #{head}"
         head + (((1 + (2 + 3))))
     end
+
+    defp accessLookup(map) do
```
HiPhish, line 26: Don't you need `x` as a function parameter?

```suggestion
    defp accessLookup(map, x) do
```

My Elixir is a bit rusty, so please double-check. The code does not have to compile, but it should if possible.

HiPhish:
Merged, thank you.

PR #77, opened 2023-12-21
feat: queries to highlight only jsx/tsx tags

ribru17:
Hello and thanks for the awesome plugin! I only use it for lisp/HTML-like languages since I find the general bracket highlights a bit distracting. But I really wanted to be able to use them for the HTML parts of JSX/TSX, so I created this PR to add a query for only the React-style tags of these files, without the main highlighting. Feel free to change this PR if needed (maybe to move from the `javascript` directory to a new `jsx` directory? I wasn't sure if maybe this would add redundant queries when using an `inherits` modeline). Thanks!

### DIFF
--- a/queries/javascript/rainbow-tags-react.scm
+++ b/queries/javascript/rainbow-tags-react.scm
@@ -1,0 +1,29 @@
+(jsx_element
+  open_tag: (jsx_opening_element
+              "<" @delimiter
+              name: (identifier) @delimiter
+              ">" @delimiter)
+  close_tag: (jsx_closing_element
+               "</" @delimiter
+               name: (identifier) @delimiter
+               ">" @delimiter @sentinel)) @container
+
+(jsx_element
+  open_tag: (jsx_opening_element
+              "<" @delimiter
+              name: (member_expression) @delimiter
+              ">" @delimiter)
+  close_tag: (jsx_closing_element
+              "</" @delimiter
+               name: (member_expression) @delimiter
+              ">" @delimiter @sentinel)) @container
+
+(jsx_self_closing_element
+  "<" @delimiter
+  name: (identifier) @delimiter
+  "/>" @delimiter @sentinel) @container
+
+(jsx_self_closing_element
+  "<" @delimiter
+  name: (member_expression) @delimiter
+  "/>" @delimiter @sentinel) @container
--- a/doc/rainbow-delimiters.txt
+++ b/doc/rainbow-delimiters.txt
@@ -353,11 +353,13 @@ addition there are the following extra queries for certain languages:
   - `rainbow-delimiters-react` Includes React support, set by default for
     Javascript files
   - `rainbow-parens` Only parentheses without React tags
+  - `rainbow-tags-react` Only React tags without parentheses
 - `query`
   - `rainbow-blocks` Highlight named nodes and identifiers in addition to
     parentheses (useful for |:InspectTree|)
 - `tsx`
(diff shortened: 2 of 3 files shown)


HiPhish:
Looks good, but I think `rainbow-tags-react` would be a better name. The `-only` suffix is redundant in my opinion, and putting `tags` before `react` fits better with `rainbow-delimiters-react`. What is your opinion?

ribru17:
Sounds good! I'll change it now

HiPhish:
One more change please: the file `queries/tsx/rainbow-tags-react.scm` should not be a symlink, use a query which inherits from Javascript instead:

```
; inherits: javascript

;;; This query exists for people who only want to highlight tags without
;;; parentheses.
```

All the other queries inherit from Typescript, whose queries in return inherit from Javascript. Since there is nothing Typescript-specific here we can inherit straight from Javascript.

After you have made the change please squash your commits and rebase on the current `master` branch.

ribru17:
Got it, should be all good now. :+1:

queries/javascript/rainbow-tags-react.scm
```diff
@@ -0,0 +1,36 @@
...
+  "/>" @delimiter @sentinel) @container
+
+(jsx_self_closing_element
+  "<" @delimiter
+  name: (member_expression) @delimiter
+  "/>" @delimiter @sentinel) @container
+
+(jsx_expression
+  "{" @delimiter
+  "}" @delimiter @sentinel) @container
```
HiPhish, line 36: Are you sure you want this pattern? It will highlight the curly braces in expressions like this as well:

```
<div style={{ display: 'flex' }}>
```

queries/javascript/rainbow-tags-react.scm
```diff
@@ -0,0 +1,36 @@
+; This query is for highlighting just the JSX tags.
+
+;;; React.js support
```
HiPhish, line 3: Please remove this comment, the file name makes it obvious that we only do React here.

queries/javascript/rainbow-tags-react.scm
ribru17: Oh thanks, good catch!

HiPhish:
Looks good now. Please squash one more time and I can merge.

ribru17:
Done. Also you may be interested in [this page](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests).

HiPhish:
Merged, thank you.

> Also you may be interested in [this page](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests).

Does that only apply to PRs merged through GitHub's web UI? I don't actually use GitHub for merging PRs, this repo is just a mirror. My workflow is to check out the PR locally, inspect it locally. Then I merge the feature branch into `master` locally and push to GitLab. GitLab then mirrors the changes to GitHub.

This way I am not tied to GitHub, but I still have the option of using it for collaboration because most people are on GitHub for better or for worse.

ribru17:
I see; I think you may be be able to do something like `git merge --squash <branchname>` but maybe at that point it would just be easier to have contributors squash beforehand.

HiPhish:
I think Git changes the committer to myself if I `git merge --squash` myself. I'll have to try it the next time to see if that is really the case though.

PR #79, opened 2023-12-30
fix: type and checkhealth updates

Danielkonge:
I did a `:checkhealth` today and noticed a problem with `rainbow-delimiters`, it seems that the health check wasn't updated with the recent updates. In this PR I updated the health file with something that should work with the recent updates.

While doing this I noticed that the types for priorities allowed the function to return nil, but I don't think we ever want that. (For strategies we allow a function return of nil to disable `rainbow-delimiters` for the given language, but for queries and priority it doesn't really make sense to allow nil. Unless you again want to use this for something?)


lua/rainbow-delimiters/health.lua
```diff
@@ -168,7 +190,15 @@ function M.check()
...
 		for lang, query in pairs(queries) do
-			if lang ~= '' then
+			if lang == '' then
+				if query ~= 'rainbow-delimiters' then
+					local msg = "Unknown custom default query"
+					warn(msg)
+				else
+					local msg = "Valid custom default query"
+					ok(msg)
+				end
```
HiPhish, line 200: I am not sure about this. If someone defines his own custom query for all the languages he uses, then that's perfectly valid. Otherwise there is no real point of having an option for `''` in the first place if only one fixed value passes the test.

lua/rainbow-delimiters/health.lua
```diff
@@ -179,6 +209,35 @@ function M.check()
...
+					local msg = string.format("Valid custom query for '%s'", lang)
+					ok(msg)
+				end
+			end
+		end
+	end
+
+	local priorities = settings.priority
+	if priorities then
+		start 'Custom pririties'
```
HiPhish, line 222: Typo

```suggestion
		start 'Custom priorities'
```

lua/rainbow-delimiters/health.lua
```diff
@@ -179,6 +209,35 @@ function M.check()
...
+		for lang, priority in pairs(priorities) do
+			local is_valid_prirority = check_priority(priority)
+			if lang == '' then
+				if is_valid_prirority then
+					local msg = "Valid custom default priority"
+					ok(msg)
+				else
+					local msg = "Invalid custom defualt priority"
```
HiPhish, line 230: Typo

```suggestion
					local msg = "Invalid custom default priority"
```

lua/rainbow-delimiters/health.lua
```diff
@@ -179,6 +209,35 @@ function M.check()
...
+					error(msg)
+				end
+			else
+				if is_valid_prirority then
+					local msg = string.format("Valid custom priority for '%s'", lang)
+					ok(msg)
+				else
+					local msg = string.format("Valid custom priority for '%s'", lang)
```
HiPhish, line 238:
```suggestion
					local msg = string.format("Invalid custom priority for '%s'", lang)
```

HiPhish:
> While doing this I noticed that the types for priorities allowed the function to return nil, but I don't think we ever want that.

You are right, that was a mistake.

lua/rainbow-delimiters/health.lua
Danielkonge: This is the reason I made it a warning instead of an error and wrote "unknown" instead of "invalid", but it might be better to just remove this. I put it there to catch spelling mistakes, but the LSP should already help with that.

Danielkonge: I tried rewriting it a bit now and return ok instead of warning. Please let me know if the updated version is fine or if you prefer to just remove this? I think it makes sense to include a check here since a lot of users probably write out the default setup from the docs.

I can squash the commits when I know what to do here.

lua/rainbow-delimiters/health.lua
```diff
@@ -168,7 +190,15 @@ function M.check()
 	if queries then
 		start 'Custom queries'
 		for lang, query in pairs(queries) do
-			if lang ~= '' then
+			if lang == '' then
+				if query ~= 'rainbow-delimiters' then
+					local msg = "User defined custom default query.\nIf you meant to use 'rainbow-delimiters', please check your setup."
```
HiPhish, line 195: I would include the name of the query, then users can check for typos right there instead of having to open their configuration. I also shortened the message a bit. What is your opinion?

```suggestion
					local msg = string.format(
						"User-defined default query '%s'\
						If you meant 'rainbow-delimiters' check for typos",
						query
					)
```

Danielkonge: This looks good to me. If the default query is setup via a function it will just print a function address, but that is probably rare, and it will still convey the correct idea to the user.

HiPhish:
Everything looks good now. Please squash and I will merge.

Danielkonge:
> Everything looks good now. Please squash and I will merge.

I have squashed the commits now.

HiPhish:
Merged, thank you.

PR #83, opened 2024-01-06
fix: java `dimensions` highlight

kawre:
The workaround i found is to treat `dimensions` node as a single delimiter which will highlight it the same color.
One minor thing is that it will also highlight any text inside dimensions the same color as the delimiter but it's not grammatically correct anyway so it's not a big problem.

### DIFF
--- a/queries/java/rainbow-delimiters.scm
+++ b/queries/java/rainbow-delimiters.scm
@@ -50,12 +50,8 @@
   "{" @delimiter
   "}" @delimiter @sentinel) @container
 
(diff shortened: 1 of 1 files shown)


HiPhish:
Oh wow, that's a really clever fix. Thank you.

> One minor thing is that it will also highlight any text inside dimensions the same color as the delimiter but it's not grammatically correct anyway so it's not a big problem.

If the code is not grammatically correct all bets are off anyway, so I'm fine with this trade-off.

PR #89, opened 2024-01-17
Implement functions for checking if rainbow is enabled in buffer

mehalter:
This just adds a nice function in the API to check if rainbow delimiters are enabled in a buffer. It is currently feasible by directly requiring and accessing the internal lib functions but that seems a bit unsafe. Also was thinking if it should also check if the value isn't `nil` to see if it's actually attached. Maybe that should be a different API function (`is_attached`)? Feedback super welcome!

Also, amazing job with this project! Very much appreciate the revival of the original archived project and all of the refactors and improvements that have gotten it to where it is today :D

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,7 @@ Unreleased
 Added
 =====
 
+- Public API function `is_enabled`
 - XML support
 - Missing patterns for Java:
    - `array_initializer`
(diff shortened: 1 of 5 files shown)


HiPhish:
Hi, sorry it took so long, but I was in the process of getting proper testing up and running in the project using Lua and Busted. And while I was at it I added a bunch of tests for this function as well and I noticed that it had a bug: if a buffer was not in the table the expression `lib.buffers[bufnr]` would evaluate to `nil`, which is unequal to `false`, and so we would get a false positive result.

Can you please squash the commits into one and rebase on the current `master`? I can merge the PR then.

mehalter:
No problem! Thanks for taking a look at this! Just squashed and rebased!

HiPhish:
Merged, thank you.

PR #91, opened 2024-01-21
fix(python): add missing tuple_pattern query

Danielkonge:
I found a missing python query.


HiPhish:
Nice catch, thank you.

PR #93, opened 2024-01-24
Add missing python and starlark queries

Danielkonge:
When I added `tuple_pattern` to python last time, I missed that it also applies to starlark.

Also, I found another missing python pattern `import_from_statement`, which doesn't seem to apply to starlark.


HiPhish:
Merged, thank you.

PR #94, opened 2024-01-25
fix(luadoc): add new luadoc queries

Danielkonge:
The luadoc parser was updated a couple of days ago, and I found two new queries to highlight in `rainbow-delimiters`.


HiPhish:
Merged, thank you.

PR #101, opened 2024-02-25
fix(fennel)!: update queries to reflect new upstream parser

reo101:
Update fennel queries to be in line with the new upstream fennel parser

https://github.com/nvim-treesitter/nvim-treesitter/pull/6132

Also add commented out query for highlighting the single `:` in table bindings


HiPhish:
Merged, thank you.

PR #104, opened 2024-03-16
feat : add matlab delimiter query

Jaehaks:
There was no delimiters query of matlab. 
Add some treesitter delimiters

### DIFF
--- a/queries/matlab/rainbow-delimiters.scm
+++ b/queries/matlab/rainbow-delimiters.scm
@@ -1,0 +1,31 @@
+(function_call
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
+
+(function_arguments
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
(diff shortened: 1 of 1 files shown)


HiPhish:
Thanks. Can you please add a test file as well?

Jaehaks:
> Thanks. Can you please add a test file as well?

Thank you for your reply, 

@HiPhish  I maked the query with some examples and tested using [garciparedes/matlab-examples](https://github.com/garciparedes/matlab-examples)
`matlab-examples/numerical-methods/*`

HiPhish:
I don't know Matlab and I don't have any need to learn it at the moment, so please add a test file yourself. The test file needs to cover all the patterns defined in the query. It would be good if it also works when executed, so that people who run a language server don't see errors and warnings everywhere, but that's not necessary. Please do not just copy-paste a file from the matlab-examples repo, they have a different license.

PR #105, opened 2024-03-21
use github yo-dawg

reaink:
error gitlab auth tls


HiPhish:
Do you always have a problem with GitLab or was it just temporary?

reaink:
Not really, it's probabilistic.

HiPhish:
> Not really, it's probabilistic.

Do you mean that it's more or less random whether the GitLab URL works for you?

I want to do all my version management with GitLab and use GitHub only as a mirror for people who prefer it, so I would like to keep the existing URL. Here is something you could try: configure Git to substitute GitLab URLs with GitHub URLs: https://stackoverflow.com/a/40435963. I have not tried it myself, but I think this could be a good compromise if it works. Please report back if it works for you.

Note also that you do not need yo-dawg.nvim to use this plugin, it's only necessary for running the automated tests.

PR #107, opened 2024-04-17
Update julia queries

mehalter:
the Julia treesitter did a huge breaking change refactor recently that is now available in `nvim-treesitter` that breaks this plugin. There is no longer a `parameter_list` and the elements that were identified with `parameter_list` are now just `argument_list`s. So we can simply remove this query.

This also adds a couple missing queries for tuple expressions and curly brace expressions

Relevant PRs:

- https://github.com/tree-sitter/tree-sitter-julia/pull/135
- https://github.com/nvim-treesitter/nvim-treesitter/pull/6429


HiPhish:
Thank you. Can you please update test test file (`test/highlight/julia/regular.jl`) as well? There are no examples of `tuple_expression` and `curly_expression`.

mehalter:
Yeah definitely @HiPhish ! I went ahead and updated the `test/highlight/julia/regular.jl` file to have examples of both a `tuple_expression` and a `curly_expression`

mehalter:
Sure thing. I just squashed the commits. I was thinking the commits were separate since one fixed a bug that was introduced when queries were removed upstream and then the other commit added in queries that were missing regardless of the changes upstream.


Thanks! Looking forward to this getting merged in :)

PR #108, opened 2024-05-02
Add support for missing R nodes

oxfist:
Thanks so much for your work! I recently migrated to this version and noticed there are missing queries to properly support R's default control structures plus function definitions. This PR adds them. 👍🏼


HiPhish:
Merged, thank you.

PR #109, opened 2024-05-07
Add a few missing javascript nodes

oxfist:
Also added some missing nodes for JavaScript and React along with examples.


HiPhish:
Thank you.

PR #117, opened 2024-05-19
Add missing queries for TypeScript interface and enum

oxfist:
Hey there! Adding a couple missing nodes for TypeScript's interface and enums.


HiPhish:
Merged, thank you.

PR #120, opened 2024-05-26
Cleanup + missing rust query + wgsl support

Danielkonge:
Neovim has updated some type signatures, so I updated them too here. 

Also `regtype = 'c'` causes errors for me, so I changed it to `regtype = 'v'`, which seems to be what we want -- see: https://neovim.io/doc/user/builtin.html#getregtype() . I am not sure when this was changed in Neovim recently, or if this code has always been wrong? (Wrong in the sense that it maybe should have been `'v'` technically, although it definitely did accept `'c'` previously.)

(Note: I haven't yet tested this in older versions of Neovim)

Finally I added a missing rust query and wgsl support in separate commits.

### DIFF
--- a/lua/rainbow-delimiters.types.lua
+++ b/lua/rainbow-delimiters.types.lua
@@ -15,7 +15,7 @@
 
 ---@class (exact) rainbow_delimiters.buffer_settings
 ---@field strategy rainbow_delimiters.strategy
----@field parser LanguageTree
+---@field parser vim.treesitter.LanguageTree
 ---@field lang string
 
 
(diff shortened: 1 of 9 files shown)


catdadcode:
I sure hope this gets merged soon. Have had to disable the plugin completely until then.

HiPhish:
@Danielkonge Can you please rebase onto the current `master`? There have been a couple of changes, including changing the `c` to `v`. That part confuses me though. According to the documentation the acceptable values are the same as in `setreg()`, which has both `c` and `v` as synonyms.

Danielkonge:
I rebased onto the current master now.

HiPhish:
Merged, thank you very much.

> is true, then we should maybe report this as an issue upstream.

It's a change in `master` over at Neovim from a few days ago. They changed the `opts` argument of `vim.highlight.range()` to where the `regtype` string conforms to `getregtype()` instead of  `setreg()`. I did not test with the nightly builds, so this went past me.

PR #125, opened 2024-06-30
`global` strategy rewrite (draft)

Danielkonge:
Since we wanted to do a rewrite to `iter_matches` anyways, I wanted to try to make it possible to write a `rainbow-blocks` query file for python with the `iter_matches` setup, but this turned out to be quite a bit harder than I thought, because `iter_matches` doesn't always give you the matches in a nice order (for what we want to do at least), but I think the current code works pretty well now. 

Do you think something like this would be too complicated? If so, I originally had a much simpler version that worked fine for the standard queries I tried it on, but it did not work at all with something like my `rainbow-blocks` python file. 

Also, I have not tried to time any of this for now, but clearly it should be slower than what we used to have. It might be an acceptable slowdown though? (I haven't noticed much difference, but I haven't checked on really large files yet.)

Some notes:
- This code uses the end of the node as sentinel if no sentinel is provided, since that was what I would want in the cases I could think of (e.g. python `rainbow-blocks`).
- I haven't cleaned up the code or added helpful comments yet. This is just my first working version with `rainbow-blocks` for python working as I would expect. 

Small update:

Although I think the code in this PR is a bit messy, I do like the resulting python block highlighting we can get with it:

Screenshot 2024-07-01 at 23 27 40

Screenshot 2024-07-01 at 23 28 34

### DIFF
--- a/lua/rainbow-delimiters.lua
+++ b/lua/rainbow-delimiters.lua
@@ -63,6 +63,7 @@ local M = {
 		['local']  = require 'rainbow-delimiters.strategy.local',
 		---Empty highlighting strategy for testing
 		['noop']   = require 'rainbow-delimiters.strategy.no-op',
+		['window'] = require 'rainbow-delimiters.strategy.window',
 	},
 	enable  = enable,
 	disable = disable,
--- a/lua/rainbow-delimiters/strategy/global.lua
+++ b/lua/rainbow-delimiters/strategy/global.lua
@@ -156,7 +156,6 @@ local function update_range(bufnr, changes, tree, lang)
 				local match_sentinel_row, match_sentinel_col = match_sentinel[1], match_sentinel[2]
 				local other_sentinel = other.sentinel ---@type integer[]
 				local other_sentinel_row, other_sentinel_col = other_sentinel[1], other_sentinel[2]
-
 				if
 					(match_start_row < other_start_row
 						or (match_start_row == other_start_row and match_start_col <= other_start_col))
@@ -229,14 +228,13 @@ local function setup_parser(bufnr, parser, start_parent_lang)
 
 				-- maybe change self_injecting_languages to { rust = true },
 				-- since that seems to be the main different one
-				local self_injecting_languages = { c = true, cpp = true, markdown = true, }
 				if not parent_lang then
 					-- If we have no parent language, then we use changes, otherwise we use the
 					-- whole tree's range.
 					-- Normalize the changes object if we have no parent language (the one we
 					-- get from on_changedtree)
 					changes = vim.tbl_map(normalize_change, changes)
-				elseif parent_lang ~= lang or self_injecting_languages[parent_lang] then
+				elseif parent_lang ~= lang or lang ~= 'rust' then
 					-- We have a parent language, so we are in an injected language code
 					-- block, thus we update all of the current code block.
 					changes = {{tree:root():range()}}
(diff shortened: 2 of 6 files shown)


Danielkonge:
Other than the `iter_matches` draft, I have also been playing around with a `window` strategy, which works fine now. I still prefer the global strategy though, since I use a fork of `indent-blankline` to highlight scope indents via the `rainbow-delimiters` extmarks, and I can't capture them with the `window` strategy, because the extmarks are ephemeral. I also have the same problem for `nvim-treesitter-context`. 

An advantage of the `window` strategy is that its performance shouldn't be influenced too much by the file size (unless there is a very large number of language injections).

HiPhish:
Hello, it has been a long time coming, but after a number of dead-ends and failed attempts I have finally finished the highlight tests. If you rebase on the current `master` you can run `make highlight-test` and Neovim will compare the current highlighting to the previously recorded highlighting. You will need Luarocks [busted](https://lunarmodules.github.io/busted/) installed for Lua 5.1. It will take a couple of seconds to run through all languages (around 13 seconds on my machine) because each tests runs in its own isolated Neovim process, but this will give results which are not affected by your personal configuration.

Here is how it works: For every language we have one or more sample files (`test/highlight/samples/<lang>/...`). An included script will open the sample files in a new Neovim instance and for each combination of file and query it records the extmarks with their range and highlight group to a "spec" file (`test/highlight/spec/<lang>/<query>/...`). A human has to check the sample file manually for correct highlighting, then he runs the script to record the extmarks.

When testing there will be a separate test generated for each combination of language and query. It fetches the extmarks like before, but instead of writing the spec file it reads the previously recorded spec and checks the given extmarks against the recorded ones. If both sets match the highlighting has not changed and the test passes.

The only downside is that since all tests are generated dynamically you cannot find them statically using something like [neotest-busted](https://github.com/HiPhish/neotest-busted). I have tagged all tests with their language and query, so you could use that to narrow down which tests to run.

With this out of the way we can now focus again on fixing the highlighting algorithm. I cannot yet promise how much time I will be able to devote to this issue, I'm really sorry about that.

Danielkonge:
> Hello, it has been a long time coming, but after a number of dead-ends and failed attempts I have finally finished the highlight tests. If you rebase on the current `master` you can run `make highlight-test` and Neovim will compare the current highlighting to the previously recorded highlighting. You will need Luarocks [busted](https://lunarmodules.github.io/busted/) installed for Lua 5.1. It will take a couple of seconds to run through all languages (around 13 seconds on my machine) because each tests runs in its own isolated Neovim process, but this will give results which are not affected by your personal configuration.
> 
> Here is how it works: For every language we have one or more sample files (`test/highlight/samples/<lang>/...`). An included script will open the sample files in a new Neovim instance and for each combination of file and query it records the extmarks with their range and highlight group to a "spec" file (`test/highlight/spec/<lang>/<query>/...`). A human has to check the sample file manually for correct highlighting, then he runs the script to record the extmarks.
> 
> When testing there will be a separate test generated for each combination of language and query. It fetches the extmarks like before, but instead of writing the spec file it reads the previously recorded spec and checks the given extmarks against the recorded ones. If both sets match the highlighting has not changed and the test passes.
> 
> The only downside is that since all tests are generated dynamically you cannot find them statically using something like [neotest-busted](https://github.com/HiPhish/neotest-busted). I have tagged all tests with their language and query, so you could use that to narrow down which tests to run.
> 
> With this out of the way we can now focus again on fixing the highlighting algorithm. I cannot yet promise how much time I will be able to devote to this issue, I'm really sorry about that.


I haven't had much time the last couple of weeks and might not have a lot of time this week either, but after that I will try to rebase on master and run the tests.

For now, you probably don't really need to look that much at this draft code. The main point is just, that it seems harder to make it work nicely with `iter_matches` than with `iter_captures`. I have an idea (and partly implemented it already) for how to maybe do this better, so it would make more sense to look at it when that is ready.

Danielkonge:
Closed in favor of #132.

PR #126, opened 2024-07-03
Update for the new R langage grammar

PMassicotte:
The parser for the R language has been recently rewritten (https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/645). This PR adapt the queries for the new node names.

### DIFF
--- a/queries/r/rainbow-delimiters.scm
+++ b/queries/r/rainbow-delimiters.scm
@@ -1,36 +1,36 @@
-(call
+(arguments
   "(" @delimiter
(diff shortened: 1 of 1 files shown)


milanglacier:
Applied this patch and it seems that it is working fine on the R file and markdown (with injection) file.

HiPhish:
Thank you. You had the `(arguments ("(" ")"))` pattern twice because `switch` became `arguments` as well. I have fixed it already.

PR #131, opened 2024-07-28
fix(c_sharp): missing tuple_pattern query

jcapblancq:
Hi,

I added a capture that was missing for tuple destructuring in C#.

Let me know if anything.

Thanks again for maintaining this project!

### DIFF
--- a/queries/c_sharp/rainbow-delimiters.scm
+++ b/queries/c_sharp/rainbow-delimiters.scm
@@ -38,6 +38,10 @@
   "(" @delimiter
   ")" @delimiter @sentinel) @container
 
+(tuple_pattern
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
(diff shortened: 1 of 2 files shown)


HiPhish:
Merged, thank you.

> Let me know if anything.

Updating the changelog and the highlight tests would be nice, but it's just a few seconds extra work for me, so it's not a big deal. See the `CONTRIBUTING` file for how to update highlight tests. These were only added a few days ago, they let me make sure that highlighting stays consistent whenever the highlighting algorithm changes, but the recorded highlights need to be updated whenever the test files change.

HiPhish:
If you have difficulty understanding my instructions feel free to open an issue. It's not a big deal if you don't update the tests, I just wanted to mention it since you asked. I have to look over your changes manually anyway before I merge them.

PR #143, opened 2024-11-07
Add missing python query `format_expression`

Danielkonge:
I found a missing python query, so this is just a minor fix.

### DIFF
--- a/queries/python/rainbow-delimiters.scm
+++ b/queries/python/rainbow-delimiters.scm
@@ -73,3 +73,7 @@
   (interpolation
     "{" @delimiter
     "}" @delimiter @sentinel) @container)
+
+(format_expression
+  "{" @delimiter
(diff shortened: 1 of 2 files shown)


HiPhish:
You have forked from quite an old state. Can you please fork off the current `master`? And while you're at it please update the sample file as well to contain a `format_expression`. I don't know what such a node looks like.

Danielkonge:
> You have forked from quite an old state. Can you please fork off the current `master`?

Ah, I accidentally pulled from my fork's master branch instead of upstream. It should be fixed now.

> And while you're at it please update the sample file as well to contain a `format_expression`. I don't know what such a node looks like.

It's been some time since I added a query, so I forgot about this. I have added it now.

HiPhish:
Oh nice, I did not know about that one. Thank you.

PR #144, opened 2024-11-09
feat: groovy support

janw4ld:
Hello! this PR adds support for `groovy` scripts (or most famously Jenkinsfiles / Pipeline DSL)

I couldn't get this working 100% because
- the TS grammar doesn't have the notion of _parenthised expressions_ which results in broken highlights for unnecessarily nested delimiters, i.e. `((1 + 2))`, see FIXME on line 22 of the test file
- there are some language constructs that are just parsed as `source_file` (which is the translation unit's root)
- some valid syntax appears as parsing errors in the AST

so LMK what you think of this patch and whether it's ok to merge with these limitations, I've been using it _in prod_ for a while :)

edit: the test file is valid groovy that can run interpreted & compile to jvm bytecode with no issue


HiPhish:
> the TS grammar doesn't have the notion of _parenthised expressions_ which results in broken highlights for unnecessarily nested delimiters, i.e. `((1 + 2))`, see FIXME on line 22 of the test file
> 
> * there are some language constructs that are just parsed as `source_file` (which is the translation unit's root)
> 
> * some valid syntax appears as parsing errors in the AST

Yes, the first point is an unfortunate problem with some parsers. Have you tried bringing the issues up with the parser maintainers? I think it's better to solve it at the source, then everyone can profit. I was able to convince the maintainers of some other parsers.

test/highlight/samples/groovy/regular.groovy
```diff
@@ -0,0 +1,71 @@
+(1..<2)/*
+    there's no special AST node for this range syntax, and the (ERROR) node here
+    isn't correct, this code is valid and runs
+*/.each { i -> // closjure
```
HiPhish, line 4: Typo
```suggestion
*/.each { i -> // closure
```

HiPhish:
The PR itself looks good to me, but the parser shortcomings are bad. What would you prefer, that we merge it as is or wait to see if the problems can be fixed upstream?

janw4ld:
I'd prefer we merge it as is for now, create an issue to the TS parser (<https://github.com/murtaza64/tree-sitter-groovy>), then update the plugin later when the parser's state improves. LMK what you think tho.

HiPhish:
OK, are you going to submit the bug report then? I don't know Groovy myself.

janw4ld:
yes, i'll gradually create issues or submit PRs for each of the bugs i listed. i started by addressing [the parenthesized expressions one](https://github.com/murtaza64/tree-sitter-groovy/issues/26) and wrote the needed [rainbow-delimiters.nvim change](https://github.com/janw4ld/rainbow-delimiters.nvim/tree/paren-support), and i'll make sure to send you a PR as soon as the parser is updated

HiPhish:
Sounds great. I have merged this PR for now.

PR #145, opened 2024-11-10
feat: added odin query

Kaikacy:
hi! this PR adds odin lang support. Hope I didn't miss anything.
#### planing to also add rainbow_blocks:
if they are possible (and not too complicated)
```odin
// switch and case matching
switch x {
    case 1:
        ...
}
```
```odin
// loop labels and break
label: for i in 0..5 {
    break label
}
```

### DIFF
--- a/queries/odin/rainbow-delimiters.scm
+++ b/queries/odin/rainbow-delimiters.scm
@@ -47,24 +47,33 @@
   "]" @delimiter @sentinel) @container
 
 (struct
-  ["{" "(" "["] @delimiter
-  ["}" ")" "]"] @delimiter @sentinel) @container
+ ("(" @delimiter
+  ")" @delimiter)?
+ ("[" @delimiter
+  "]" @delimiter)?
+ ("{" @delimiter
+  "}" @delimiter @sentinel)
+  ) @container
 
 (map_type
   "[" @delimiter
   "]" @delimiter @sentinel) @container
 
 (map
-  ["{" "["] @delimiter
-  ["}" "]"] @delimiter @sentinel) @container
+  "[" @delimiter
+  "]" @delimiter
+  "{" @delimiter
+  "}" @delimiter @sentinel) @container
 
 (bit_set_type
   "[" @delimiter
   "]" @delimiter @sentinel) @container
 
 (bit_set
-  ["{" "["] @delimiter
-  ["}" "]"] @delimiter @sentinel) @container
+  "[" @delimiter
+  "]" @delimiter
+  "{" @delimiter
+  "}" @delimiter @sentinel) @container
 
 (struct_declaration
   "{" @delimiter
--- a/queries/odin/rainbow-delimiters.scm
+++ b/queries/odin/rainbow-delimiters.scm
@@ -30,7 +30,15 @@
   "[" @delimiter
   "]" @delimiter @sentinel) @container
 
+(slice_expression
+  "[" @delimiter
+  "]" @delimiter @sentinel) @container
+
 (block
+  "{" @delimiter
+  "}" @delimiter @sentinel) @container
+
+(switch_statement
   "{" @delimiter
   "}" @delimiter @sentinel) @container
 
@@ -39,40 +47,24 @@
   "]" @delimiter @sentinel) @container
 
 (struct
-  "{" @delimiter
-  "}" @delimiter @sentinel) @container
-
-(struct
-  "(" @delimiter
-  ")" @delimiter @sentinel) @container
-
-(struct
-  "[" @delimiter
-  "]" @delimiter @sentinel) @container
+  ["{" "(" "["] @delimiter
+  ["}" ")" "]"] @delimiter @sentinel) @container
 
 (map_type
   "[" @delimiter
   "]" @delimiter @sentinel) @container
 
 (map
-  "{" @delimiter
-  "}" @delimiter @sentinel) @container
-
-(map
-  "[" @delimiter
-  "]" @delimiter @sentinel) @container
+  ["{" "["] @delimiter
+  ["}" "]"] @delimiter @sentinel) @container
 
 (bit_set_type
   "[" @delimiter
   "]" @delimiter @sentinel) @container
 
 (bit_set
-  "{" @delimiter
-  "}" @delimiter @sentinel) @container
-
-(bit_set
-  "[" @delimiter
-  "]" @delimiter @sentinel) @container
+  ["{" "["] @delimiter
+  ["}" "]"] @delimiter @sentinel) @container
 
 (struct_declaration
   "{" @delimiter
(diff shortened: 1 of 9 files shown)


queries/odin/rainbow-delimiters.scm
```diff
@@ -0,0 +1,91 @@
+(parameters
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
+
+(tuple_type
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
```
HiPhish, line 7: There is no example of this pattern.

queries/odin/rainbow-delimiters.scm
```diff
@@ -0,0 +1,91 @@
...
+
+(parenthesized_expression
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
```
HiPhish, line 11: There is no example of this pattern.

queries/odin/rainbow-delimiters.scm
```diff
@@ -0,0 +1,91 @@
...
+
+(polymorphic_type
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
+
+(attribute
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
```
HiPhish, line 23: I don't know Odin, can `attribute` ever be nested within other delimiters or can it contain nested delimiters, or is it always a top-level element? If it's the latter I don't know if it should really be included in the query. What's your opinion?

queries/odin/rainbow-delimiters.scm
```diff
@@ -0,0 +1,91 @@
...
+  "{" @delimiter
+  "}" @delimiter @sentinel) @container
+
+(struct
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
+
+(struct
+  "[" @delimiter
+  "]" @delimiter @sentinel) @container
```
HiPhish, line 51: Could these three pattern be merged into one? Their delimiters are all on the same level, so they should have the same highlighting. You might have to make some of them optional though.

queries/odin/rainbow-delimiters.scm
```diff
@@ -0,0 +1,91 @@
...
+
+(map_type
+  "[" @delimiter
+  "]" @delimiter @sentinel) @container
```
HiPhish, line 55: There is no example of this pattern.

queries/odin/rainbow-delimiters.scm
```diff
@@ -0,0 +1,91 @@
...
+
+(map
+  "{" @delimiter
+  "}" @delimiter @sentinel) @container
+
+(map
+  "[" @delimiter
+  "]" @delimiter @sentinel) @container
```
HiPhish, line 63: Same as with `struct`, can you merge these patterns into one?

queries/odin/rainbow-delimiters.scm
```diff
@@ -0,0 +1,91 @@
...
+
+(bit_set
+  "{" @delimiter
+  "}" @delimiter @sentinel) @container
+
+(bit_set
+  "[" @delimiter
+  "]" @delimiter @sentinel) @container
```
HiPhish, line 75: There is no example with curly braces. Can you merge the two patterns into one?

queries/odin/rainbow-delimiters.scm
HiPhish: Even if it does not contain other delimiters, but can be contained inside other delimiters that's reason enough to include it.

> there are some examples where it contains other delimiters

I think that's good enough of a reason even if it is rare.

queries/odin/rainbow-delimiters.scm
```diff
@@ -0,0 +1,83 @@
...
+  "{" @delimiter
+  "}" @delimiter @sentinel) @container
+
+(array_type
+  "[" @delimiter
+  "]" @delimiter @sentinel) @container
+
+(struct
+  ["{" "(" "["] @delimiter
+  ["}" ")" "]"] @delimiter @sentinel) @container
```
HiPhish, line 51: This does not highlight correctly. How about this?

```suggestion
(struct
  ("(" @delimiter
   ")" @delimiter)?
  ("[" @delimiter
   "]" @delimiter)?
  "{" @delimiter
  "}" @delimiter @sentinel) @container

```

It assumes that the curly braces are mandatory and the parentheses and square brackets are optional.

queries/odin/rainbow-delimiters.scm
```diff
@@ -0,0 +1,83 @@
...
+
+(map_type
+  "[" @delimiter
+  "]" @delimiter @sentinel) @container
+
+(map
+  ["{" "["] @delimiter
+  ["}" "]"] @delimiter @sentinel) @container
```
HiPhish, line 59: Same problem as with `struct`. How about this?

```suggestion(map
  "[" @delimiter
  "]" @delimiter
  "{" @delimiter
  "}" @delimiter @sentinel) @container
```

The only `map` example in your code has all delimiters. Is it possible in Odin to have only of of the two sets?

queries/odin/rainbow-delimiters.scm
```diff
@@ -0,0 +1,83 @@
...
+
+(bit_set_type
+  "[" @delimiter
+  "]" @delimiter @sentinel) @container
+
+(bit_set
+  ["{" "["] @delimiter
+  ["}" "]"] @delimiter @sentinel) @container
```
HiPhish, line 67: Same as with `map`.

```suggestion
(bit_set
  "[" @delimiter
  "]" @delimiter
  "{" @delimiter
  "}" @delimiter @sentinel) @container
```

Kaikacy:
I have no idea how to write these files. I just looked at some examples for other languages and docs. still don't understand what does optional and sentinel mean. don't explain it, I'm not going to write queries for any other language. without `?` it says Incorrect pattern, but it works as expected now.

HiPhish:
> don't explain it, I'm not going to write queries for any other language.

Too late, you can't stop me now ;)

> still don't understand what does optional and sentinel mean

`@sentinel` is a remainder of an earlier version of this plugin, it will be removed eventually. It marks the last delimiter of the pattern and it was an ugly but necessary hack. In programming a "sentinel" is an element of a sequence that marks the end of that sequence. For example in C there is no real string type, we say that a string is "an array of characters where the last element is the NUL character". So even if we allocate enough memory for twenty characters, if the fifth one is NUL, then then string is four characters long. In most language these days there is no need for sentinels because the length of a sequence is well-defined.

The question mark is a quantifier, it means there may be zero of one instances of that pattern. In this case is means for example there may or may not be *one* pair of parentheses inside a `struct` node.

Kaikacy:
Thanks a lot

PR #147, opened 2024-11-21
Added init.lua

Muizzyranking:
Adding this file would allow users call setup `require("rainbow-delimiters").setup()` instead of 
`require("rainbow-delimiters.setup").setup()` as the first is what I've seen many plugins authors do

Another reason is so lazy.nvim users can use opts instead of config which i think is recommended by lazy.nvim

### DIFF
--- a/lua/rainbow-delimiters/init.lua
+++ b/lua/rainbow-delimiters/init.lua
@@ -1,0 +1,13 @@
+local M = {}
+
+---Apply the given configuration to the rainbow-delimiter settings.  Will
+---overwrite existing settings.
+---
+---@param opts rainbow_delimiters.config  Settings, same format as `vim.g.rainbow_delimiters`
+function M.setup(opts)
+	vim.g.rainbow_delimiters = opts
+end
+
+return M
+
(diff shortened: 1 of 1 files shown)


Muizzyranking:
this is an example.

```lua
return {
  "HiPhish/rainbow-delimiters.nvim",
  event = "BufRead",
  opts = {
    enable_when = function(bufnr)
      local max_filesize = 1 * 1024 * 1024 -- 1 MB
      local ok, stats = pcall(vim.loop.fs_stat, vim.api.nvim_buf_get_name(bufnr))
      if ok and stats and stats.size > max_filesize then
        return nil
      end
    end,
  },
  -- config = function()
  --   require("rainbow-delimiters.setup").setup({
  --     enable_when = function(bufnr)
  --       local max_filesize = 1 * 1024 * 1024 -- 1 MB
  --       local ok, stats = pcall(vim.loop.fs_stat, vim.api.nvim_buf_get_name(bufnr))
  --       if ok and stats and stats.size > max_filesize then
  --         return nil
  --       end
  --     end,
  --   })
  -- end,
}
```

PR #151, opened 2024-12-27
java: add query for switch expression body

janw4ld:
This PR fixes Java 13's switch expressions having no highlights for their body's curly braces.   
image


HiPhish:
Thank you. Next time please add an example of the pattern to the sample file as well (test/highlight/samples/<language>). It's not a big deal here though because I know enough Java to do it myself.

PR #152, opened 2024-12-28
fix: Add enum list to cpp query

Isrothy:


HiPhish:
Thank you. Next time please add an example of the pattern to the sample file as well (`test/highlight/samples/<language>`). It's not a big deal here though because I know enough C++ to do it myself.

PR #154, opened 2025-01-05
fix: calculate priority using vim.hl.priorities

ribru17:
Hello! In the future, it is likely that the default treesitter highlight priority will change (see https://github.com/neovim/neovim/pull/31853), which breaks this plugin. It would be nice to calculate these using `vim.highlight.priorities` (`vim.hl.priorities` on nightly). Although I am unsure how to calculate this in the vimscript example, so some help would be appreciated :) Thanks!

### DIFF
--- a/lua/rainbow-delimiters/default.lua
+++ b/lua/rainbow-delimiters/default.lua
@@ -16,6 +16,7 @@
 
 ---Default plugin configuration.
 ---@type rainbow_delimiters.config
+local priorities = (vim.hl or vim.highlight).priorities
 local M = {
 	---Query names by file type
 	query = {
@@ -27,7 +28,7 @@ local M = {
 		[''] = require 'rainbow-delimiters.strategy.global',
 	},
 	priority = {
-		[''] = 110,
+		[''] = (priorities.treesitter + priorities.semantic_tokens) // 2
 	},
 	---Event logging settings
 	log = {


doc/rainbow-delimiters.txt
```diff
@@ -101,6 +101,7 @@ take precedence.
 
 Here is an example configuration:
 >vim
+    " TODO: Calculate priority using vim.(hl or highlight).priorities
```
HiPhish, line 104: I don't think we need to include a dynamically computed value in the example. The point of the sample is "here is an entry and it's a number", everything else is just complicating things. What's your opinion?

lua/rainbow-delimiters/default.lua
```diff
@@ -27,7 +27,7 @@ local M = {
 		[''] = require 'rainbow-delimiters.strategy.global',
 	},
 	priority = {
-		[''] = 110,
+		[''] = (vim.hl or vim.highlight).priorities.treesitter + 100,
```
HiPhish, line 30: Adding 100 to the `treesitter` priority would put us at 200 on Neovim 0.10.3, which is the `user` priority. How about instead we compute a value somewhere in-between `treesitter` and `semantic_tokens` based on whatever the current values are?

```suggestion
		local priorities = (vim.hl or vim.highlight).priorities
		[''] = vim.fn.round(priorities.semantic_tokens / 2 - priorities.treesitter / 2)
```

doc/news.txt
```diff
@@ -387,8 +387,10 @@ Missing patterns:
 ------------------------------------------------------------------------------
 CHANGED
 
-Default highlight priority is 110 instead of 210, which is between Tree-sitter
-and LSP semantic tokens
+• Default highlight priority is 110 instead of 210, which is between Tree-sitter
+  and LSP semantic tokens
+• Default highlight priority is now calculated from
+  `vim.(hl or highlight).priorities`
```
HiPhish, line 393: New changes belong to the top of the changelog. You can omit this if you want to, I can fill in the changelog myself.

doc/rainbow-delimiters.txt
ribru17: Sure, that makes sense

lua/rainbow-delimiters/default.lua
ribru17: Nice, that's smart

doc/news.txt
ribru17: Thanks, I'll omit then

lua/rainbow-delimiters/default.lua
HiPhish: Sorry, you are right. The formula is:

```
t + (s - t) / 2 = t + s/2 - t/2 = s/2 + t/2
```

lua/rainbow-delimiters/default.lua
```diff
@@ -27,7 +28,7 @@ local M = {
 		[''] = require 'rainbow-delimiters.strategy.global',
 	},
 	priority = {
-		[''] = 110,
+		[''] = (priorities.treesitter + priorities.semantic_tokens) // 2
```
HiPhish, line 31: FYI: Lua 5.1 does not support the `//` operator, it was introduced in Lua 5.3.

HiPhish:
Merged, thank you.

PR #159, opened 2025-02-23
Update treesitter query language queries

ribru17:
A new missing node syntax was added, and this commit adds support for that.

### DIFF
--- a/doc/news.txt
+++ b/doc/news.txt
@@ -33,6 +33,8 @@ Missing patterns:
         - `enumerator_list`
     • Java:
         - `switch_block`
+    • Query:
+        - `missing_node`
 
 ------------------------------------------------------------------------------
 CHANGED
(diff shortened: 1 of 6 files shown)


HiPhish:
Can you please provide an example for that?

ribru17:
## Before
image


## After
image

Note that the non-rainbow blocks version is the exact same, just without the node name highlight (like all the others)

HiPhish:
I think the `MISSING` should be highlighted the same as the delimiters in `rainbow-blocks` because it is a fixed part of the `missing_node` just like the parentheses themselves.

Screenshot_20250301_090937

Screenshot_20250301_090919

What's your opinion?

ribru17:
Sounds good, it looks nicer this way👍

HiPhish:
Merged, thank you.

PR #171, opened 2025-04-16
feat: add support for nushell

blindFS:
Homepage of the shell language: https://www.nushell.sh

image

Here I intentionally overlooked some delimiters where nested structures are not allowed by the grammar.


HiPhish:
Merged, thank you.

> Here I intentionally overlooked some delimiters where nested structures are not allowed by the grammar.

Sounds reasonable for the time being. It is possible for users to override the default if they want to.

PR #179, opened 2025-06-11
add glsl support, fix cpp and latex rainbow-blocks

sevenninesixthree:
I noticed that in cpp, for loop like `for(auto &a:as)a=1;` and lambda capture like `[h]` in `auto f=[h]{return h;};` have no highlight, so I fixed it. In latex, I found that under rainbow-blocks, every `{}` highlights such as `\frac{}{}` which were work well were disappear, it now is fixed, and the support of highlighting `\left` `\right` pair is added in rainbow-blocks.

### DIFF
--- a/queries/latex/rainbow-blocks.scm
+++ b/queries/latex/rainbow-blocks.scm
@@ -2,6 +2,10 @@
   "$" @delimiter
   "$" @delimiter @sentinel) @container
 
+(inline_formula
+  "[" @delimiter
+  "]" @delimiter @sentinel) @container
+
 (generic_environment
   (begin) @delimiter
   (end)   @delimiter @sentinel) @container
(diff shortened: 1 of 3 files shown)


HiPhish:
Thank you for the PR. Can you please add samples for the patterns? You can add them to the sample files in `test/highlight/samples/<language>`. I have added them for C++, but I don't know GLSL.

sevenninesixthree:
I have added them for glsl, and have fixed some other missing situation for cpp (`auto c = new int [10]`) and latex (`\author{}` and `\left(...\right)`)

queries/latex/rainbow-blocks.scm
```diff
@@ -9,3 +9,39 @@
...
+
+(math_delimiter
+  left_command: _ @delimiter
+  left_delimiter: _ @delimiter @sentinel
+  right_command: _ @delimiter
+  right_delimiter: _ @delimiter @sentinel) @container
+
+(brack_group_key_value
+  "[" @delimiter
+  "]" @delimiter @sentinel) @container
```
HiPhish, line 21: There is no example of this pattern.

HiPhish (changes requested):
There are two `ERROR` nodes in your GLSL sample file.

queries/latex/rainbow-blocks.scm
sevenninesixthree: The first `ERROR` node is caused by how nvim parse macro, and this `ERROR` node will pop out in any language that has macro, such as C, which will not affect the highlights. I delete the statement that produce the second `ERROR` node. The missing test now is added in latex.

HiPhish:
Thank you, I have merged it. I prefer separate PRs for different language, it makes things easier to review and fixes for one language do not prevent fixes to the other languages from being merged when they are done.

Issue #177, opened 2025-06-02
[Bug]: rainbow go away when scrolling cpp file


sevenninesixthree:
### Neovim version

0.11.1

### Language affected

c++

### Strategy

global

### Description

same as #174 but now in c/c++. When tree-sitter is enable and a used macro is scrolled out of the screen, rainbow highlights will disappear. An example:
```c
#define U 10
int main(){
  int u=U;
  \\ blanks
  \\ ...
  return 0;
}
```
My config, 
lazy.nvim
```lua
require("lazy").setup{
  {
  'nvim-treesitter/nvim-treesitter',
  lazy = false,
  branch = 'main',
  build = ':TSUpdate'
  },{"hiphish/rainbow-delimiters.nvim"}}
```
treesitter & rainbow-delimiters
```lua
vim.api.nvim_create_autocmd({'BufEnter','BufWinEnter'}, {
  pattern = { '*.lua','*.cpp','*.c','*.h','*.tex','*.glsl','*.sh'},
  callback = function() vim.treesitter.start() end,
})
vim.g.rainbow_delimiters = {
  strategy = {
    [''] = 'rainbow-delimiters.strategy.global',
    vim = 'rainbow-delimiters.strategy.local',
  },
  query = {
    [''] = 'rainbow-delimiters',
    lua = 'rainbow-blocks',
    latex = 'rainbow-blocks'
  },
  priority = {
    [''] = 110,
    lua = 210,
  },
  highlight = {
    'RainbowDelimiterRed',
    'RainbowDelimiterYellow',
    'RainbowDelimiterBlue',
    'RainbowDelimiterOrange',
    'RainbowDelimiterGreen',
    'RainbowDelimiterViolet',
    'RainbowDelimiterCyan',
  },
  log={level=vim.log.levels.DEBUG}
}
```

sevenninesixthree:
This example that do not produce error node will also cause that problem,
```c
#define FOO return 0
int main(){
/*blank lines*/
  FOO;
}
```
so it is actually an issue caused by this plug.

Issue #176, opened 2025-05-20
[Bug]: The plugin has failed


XuanseXS:
### Neovim version

0.11.1

### Description

I have encountered a bug. When I open a C file, the plugin is working, but when I move the cursor, the plugin seems to be malfunctioning. Do you have any troubleshooting ideas?

When the file is first opened
![Image](https://github.com/user-attachments/assets/da335429-cfe6-4bac-a62d-026b8338137d)

After moving the cursor

![Image](https://github.com/user-attachments/assets/8779b48f-c486-4f22-8828-b9c31b4fa891)

XuanseXS:
```
    {
        'nvim-treesitter/nvim-treesitter',
        run = ':TSUpdate',
        config = function()
            require('nvim-treesitter.configs').setup {
                ensure_installed = { "c", "lua", "make", "cpp", "python", "bash" },
                highlight = {
                    enable = true,
                    additional_vim_regex_highlighting = false
                },
            }
        end,
    },

    {
        'HiPhish/rainbow-delimiters.nvim',
        version = "v0.9.0",
        lazy = false,
        config = function()
            require "configs.highlight"
        end,
    }
```
This is my plugin configuration

realhackcraft:
I have something similar, but I can move the cursor for a while before the rainbow is gone. And for me, it only happens with moving the viewport around. (So zz also causes this). Looks like it can be fixed with `:edit`

realhackcraft:
I just investigated a bit and for my case, it was because the tree-sitter create ERROR nodes with rust macros. In your cases, it might be different, but you can check if ERROR nodes are causing the problem with `:InspectTree`.

XuanseXS:
@realhackcraft The solution works great. Rainbow delimiters now show up properly in C files.

realhackcraft:
> [@realhackcraft](https://github.com/realhackcraft) The solution works great. Rainbow delimiters now show up properly in C files.

Keep in mind it's temporary and probably will happen again next time you open the file. See #2 for more automated hacks.

psi4j:
> I encountered the same issue in a C++ file where rainbow delimiters disappear when scrolling.
> 
> As a workaround, I added the following autocommand to re-enable the plugin on WinScrolled, and it seems to work well:
> 
> vim.api.nvim_create_autocmd("WinScrolled", {
>   callback = function()
>     vim.schedule(function()
>       require("rainbow-delimiters").enable(0)
>     end)
[...]

Issue #174, opened 2025-04-17
[Bug]: Highlights go away when scrolling files


IDontKnow11010111000:
### Neovim version

0.11.0

### Language affected

rust

### Strategy

global

### Description

When scrolling through rust files, the highlights disappears (returns to same color for all) and only returns when the file is reoppened/:edit-ed.  This happens when moving with <c-E/Y/D/U> and also just j and k.

It happens when moving from where `panic!()` i the top line past where the line below it (line 3 at top of window to after line 4), if there is nothing but white on the rest of the visible buffer.
```rust
fn foo() {
    {
        panic!()
    }
}
```

The correspond log is:

```
2025-04-17T00:12-0500	DEBUG	strategy.global.current	Setting up parser for buffer 3
2025-04-17T00:12-0500	DEBUG	strategy.global.current	Setting up parser for 'rust' in buffer 3
2025-04-17T00:12-0500	DEBUG	strategy.global.current	Performing full updated on buffer 3
2025-04-17T00:12-0500	DEBUG	strategy.global.current	Updated range with changes { { 0, 0, 4294967295, 4294967295 } }
2025-04-17T00:12-0500	DEBUG	strategy.global.current	Setting up parser for buffer 3
2025-04-17T00:12-0500	DEBUG	strategy.global.current	Setting up parser for 'rust' in buffer 3
2025-04-17T00:12-0500	DEBUG	strategy.global.current	Performing full updated on buffer 3
2025-04-17T00:12-0500	DEBUG	strategy.global.current	Updated range with changes { { 0, 0, 4294967295, 4294967295 } }
2025-04-17T00:12-0500	DEBUG	strategy.global.current	Setting up parser for buffer 3
2025-04-17T00:12-0500	DEBUG	strategy.global.current	Setting up parser for 'rust' in buffer 3
2025-04-17T00:12-0500	DEBUG	strategy.global.current	Performing full updated on buffer 3
```

(while in the file with broken highlighting, :edit, then mode down with <c-E>).

The most minimal version of my init.vim I have bothered to test and can replicate it with is (I doubt the colorscheme is affecting anything):
```lua
    require('lazy').setup({
        spec = {
            {
                'nvim-treesitter/nvim-treesitter',
                lazy = false,
                build = ':TSUpdate'
            },
            {
                'morhetz/gruvbox',
                lazy = false
            },
            {
                'hiphish/rainbow-delimiters.nvim',
[...]

PR #182, opened 2025-07-15
fix that language have macro (rust, c, cpp...) will unexpectly clear namespace

sevenninesixthree:
See #176 #177 and #174.


HiPhish:
I have taken a quick look at your PR and it does appear to fix the issue. I'll have to look into it more and see if I can write a test before I can merge. Thank you very much for investigating the issue.

HiPhish:
Thank you very much, I have added a new test and all the existing tests pass. I had never seen this issue myself since I rarely use any of the affected language these days, at least not in a file that's large enough to reproduce the problem.

PR #186, opened 2025-07-24
python: Add missing `with_clause`

Danielkonge:
Noticed that this was missing while randomly using a long with clause.


HiPhish:
Nice catch, thank you.

PR #188, opened 2025-08-01
Add some missing queries for OCaml

S41d:


queries/ocaml/rainbow-blocks.scm
```diff
@@ -2,13 +2,35 @@
 (let_expression ; Line 170
   (value_definition
     "let" @delimiter
-    ("rec" @delimiter)?)
+    ("rec" @delimiter)?
```
HiPhish, line 5: I cannot find any examples of `"rec"` in the sample file. I know this is not your fault, but can you please add one? I don't know OCaml myself.

queries/ocaml/rainbow-blocks.scm
```diff
@@ -2,13 +2,35 @@
...
+    ("and" @delimiter (let_binding))*)
```
HiPhish, line 6: Do we really need the `(let_binding)` after the `"and"`?

queries/ocaml/rainbow-blocks.scm
```diff
@@ -2,13 +2,35 @@
...
+  "in" @delimiter @sentinel) @container
+
+(let_expression ; Line 256
+  (value_definition
+    (let_operator) @delimiter
+    ("rec" @delimiter)?
+    ((let_and_operator) @delimiter (let_binding))*)
+  "in" @delimiter @sentinel) @container
```
HiPhish, line 14: There is only one example of `(let_operator)`, but it has no `"rec"`  node. Can you please add an example?

HiPhish (changes requested):

queries/ocaml/rainbow-blocks.scm
S41d: Of course, no problem.

queries/ocaml/rainbow-blocks.scm
S41d: I just copy pasted the query for `let_expression`  and forgot to remove the rec part, mb 😅. I'll fix it right away

queries/ocaml/rainbow-blocks.scm
HiPhish: I see. We could use the wildcard `_` (or `(_)` if we only want to match named nodes) instead of `(let_binding)`. Would that make more sense or would it be too generic?

HiPhish:
If it's OCaml you can add it to this PR, if it's a different language please open a new one.

S41d:
Yes I meant to commit the OCaml ones only, sorry 😅.

queries/ocaml/rainbow-blocks.scm
```diff
@@ -173,3 +201,24 @@
...
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
+
+(type_binding ; Line 276
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
+
+(parameter ; Line 252
+  (("?") @delimiter)?
+  (("~") @delimiter)?
```
HiPhish, line 218: I think we can omit the inner parentheses.

```suggestion
  ("?" @delimiter)?
  ("~" @delimiter)?
```

queries/ocaml/rainbow-delimiters.scm
```diff
@@ -140,3 +143,24 @@
...
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
+
+(type_binding ; Line 276
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
+
+(parameter; Line 252
+  (("?") @delimiter)?
+  (("~") @delimiter)?
```
HiPhish, line 160: Same as above.

```suggestion
  ("?" @delimiter)?
  ("~" @delimiter)?
```

Issue #185, opened 2025-07-21
github actions checkout step fails for repos with rainbow-delimiters.nvim as a submodule


perrin4869:
I have setup a test action for my dotfiles, which include this plugin as a submodule, and after 49372aa the actions started failing at the checkout step:

https://github.com/perrin4869/dotfiles/actions/runs/16405685718/job/46352503720?pr=5263

jergason:
Also seeing failures when running `:PackerUpdate` with this plugin:

```
 ✗ Failed to update HiPhish/rainbow-delimiters.nvim
  Errors:
    Already up to date.
    fatal: No url found for submodule path '.luals/addons/busted' in .gitmodules
    49372aa
```

HiPhish:
Oh dear, it's impossible to please everyone. I have renamed the `.gitmodules` file to `.submodules` and now you are getting errors because your plugin managers expect the file to be there.

Here is what is going on: I have a couple of dependencies for testing which are included via Git submodules. These dependencies are only for development, if you install them as a user they are just bloat. The problem is that plugin managers would install those submodules by default unless told not to. I figured it would be better if this plugin were to play nice with plugin managers out of the box, so I renamed the file. But now it looks like this is causing problems for people who already had the plugin installed.

What should I do? Revert back at the cost of future users, or keep things as they are at the cost of existing users?

perrin4869:
Please note that in my case, it is not the plugin manager that is failing, since I am effectively not using any plugin manager.
I think it's clear that git behavior when you setup a submodule without a .gitmodules file present is not clearly defined, and probably not officially supported. It's probably advisable to try another approach to setting up these dev dependencies?

HiPhish:
@perrin4869 The idea is to not set up Git submodules in the first place, unless you want to run the tests. If you do want to run tests (which I don't know why you would unless you want to contribute code) you will have to symlink `.submodules` to `.gitmodules` and then do `git submodule init && git submodule update --checkout` explicitly. This was by design, it inconveniences contributors slightly by having them put in two lines, but things should just work for normal users.

I your case I think the solution could be to wipe this plugin from your configuration and then re-add it. That should get rid of the submodules.

perrin4869:
I tried that, you can see it in my commit, it didn't fix it. The checkout action will fail no matter what.
I guess if the plugin manager uses a git clone to checkout the plugin, that would work indeed.
[...]

PR #190, opened 2025-09-26
Revert "Hide Git submodules from plugin managers"

perrin4869:
Fixes https://github.com/HiPhish/rainbow-delimiters.nvim/issues/185

This PR reverts commit 49372aadaaf04d14a50efaa34150c51d5a8047e1 which introduced a regression on workflows that used this repository as a git submodule.

I understand the motivation behind the original change, but until we have a forward-compatible fix, reverting is the least risky option to restore stability for affected users.

This revert is minimal and self-contained. If you’d prefer a different direction, I’m happy to rework this PR, but I wanted to provide a quick path to unblock users.


HiPhish:
Can we close this after a78de984583e4d0fbab9f8532c5b21e0aede55b8?

PR #194, opened 2025-10-16
feat: add support for jinja templates (htmljinja)

tjkeller-xyz:
This adds the same file for htmljinja as is included for htmldjango. There are also the jinja and jinja2 filetypes that usually include HTML, but I always just write `{# vim: set filetype=htmljinja: #}` at the top of my files since vim usually picks it up as `htmldjango` filetype otherwise.

### DIFF
--- a/queries/htmljinja/rainbow-delimiters.scm
+++ b/queries/htmljinja/rainbow-delimiters.scm
@@ -1,0 +1,3 @@
+;;; Intentionally empty.  A Jinja template contains HTML, and this file exists
+;;; so that the HTML nodes can be highlighted.  This query only exists to
+;;; satisfy the requirements of this plugin.


HiPhish:
Does this actually work? Nvim-treesitter has support for `jinja` and `jinja_inline` on both `master` and `main`, but not `htmljinja`.

tjkeller-xyz:
It works perfectly in conjunction with [geigerzaehler/tree-sitter-jinja2](https://github.com/geigerzaehler/tree-sitter-jinja2). I think the only reason that upstream lacks support is that `htmljinja` files are almost always misinterpreted by nvim as `htmldjando` so people dont often run into the issue

HiPhish:
OK, that makes sense. Upstream has support for plain `jinja`, would it maybe make more sense to support that instead? The difference between Jinja and Django templates is that Jinja is not limited to HTML, it can work with any other language. You could for example use Jinja to generate LaTeX code to produce a PDF document.

tjkeller-xyz:
Unfortunately, the plain jinja filetype, as far as I can tell, doesnt support highlighting html markup in treesitter. If the markup is not highlighted, that defeats the purpose of rainbow delimiters for this filetype.

E.g.

with `htmljinja` (using this PR):
image

with `jinja`:
image

Also, apologies for making you look at this ugly markup lol

HiPhish:
You are right, I checked the regular `jinja` parser and it's pretty much useless for rainbow delimiters because there is no hierarchical structure. That means that in a template like this:

```
{% for item in items %}
{% if item > 0 %}
<p>{{ item}} is greater than zero</p>
{% endif %}
{% endfor %}
```

all the statements are on the same level, the node tree does not mark the `if` being inside the `for`. As a little aside, I see you have a non-standard HTML tag (`sl-contextmenu`) in your code; are you using some component library for Jinja? I have been using [JinjaX](https://jinjax.scaletti.dev/) and it works well enough, but it has some quirks, so I'm curious if there is something better out there.

tjkeller-xyz:
Oh interesting I haven't heard of JinjaX. Actually it looks like it might help out a bit in my current project too since as you can see in that screenshot I use a lot of jinja macros. 

The `sl-contextmenu` is a custom element created using the native [webcomponents api](https://developer.mozilla.org/en-US/docs/Web/API/Web_components). With it, you can declare custom elements (which must include a hyphen in the name of the element, e.g. `<sl-...>`) that have their own isolated shadow root, slots, and even element internals for creating custom form elements. Then you can include the custom elements in your markup. They're natively supported in every browser for many years now, but it's a little known feature since most devs are just using a frontend framework nowadays in lieu of the native components.

There are also some frameworks for creating webcomponents like Lit and more recently Svelte actually now allows you to compile any svelte component into a native webcomponent (which is my new favorite method).

I use it primarily for any element that needs a lot of its own interactive functionality, so a lot of custom form elements, special element containers, or anything element that I want to have some encapsulated functionality on the browser. But as you can see, I try to do as much serverside rendering of the static markup as possible.

Hope that helps, I definitely recommend trying it out they are really great for all kinds of things!

HiPhish:
Looks like the custom parser does not work with the `main` branch of nvim-treesitter: https://github.com/geigerzaehler/tree-sitter-jinja2/issues/5. I would like to put this issue on hold until there is a definitive way of installing the parser because I want to add tests.

> The `sl-contextmenu` is a custom element created using the native [webcomponents api](https://developer.mozilla.org/en-US/docs/Web/API/Web_components).

Ah, thanks. I knew already about web components, but I too prefer to do as much rendering as possible on the server. For small reactive stuff I use [Alpine.js](https://alpinejs.dev/) and for big stuff I use React because one of the big libraries is React only, so I don't really have a use-case for web components yet. But yeah, it's cool that they exist.

tjkeller-xyz:
Sounds good I wasn't aware of that issue. It seems like the maintainer there is fairly active so hopefully it will be resolved quickly and this can merge. Thanks for your work!

tjkeller-xyz:
https://github.com/geigerzaehler/tree-sitter-jinja2/issues/5#issuecomment-5001983551
Im able to get this working now on nvim 0.12.4! Please let me know if you get a chance to try out the updated parser and if we can merge support for it here

HiPhish:
Hi again, I was not able to get it to work because you called the language `htmljinja`, but nvim-treesitter does not support that language. It supports `jinja` and `jinja_inline` though: https://github.com/nvim-treesitter/nvim-treesitter/blob/main/SUPPORTED_LANGUAGES.md

I guess you have manually installed this parser: https://github.com/geigerzaehler/tree-sitter-jinja2/

I would prefer to stick to parsers supported by nvim-treesitter, but that's not a hard line. Have you tried the two parsers nvim-treesitter already supports? And if those don't work for your needs, how did you install the `htmljinja` parser?

tjkeller-xyz:
> Unfortunately, the plain jinja filetype, as far as I can tell, doesnt support highlighting html markup in treesitter. If the markup is not highlighted, that defeats the purpose of rainbow delimiters for this filetype.
> 
> E.g.
> 
> with `htmljinja` (using this PR): image
> 
> with `jinja`: image

Yes I installed it via the https://github.com/geigerzaehler/tree-sitter-jinja2 plugin

HiPhish:
I need to generate tests for `htmljinja`, so I have pushed two more commits: one to add the `geigerzaehler/tree-sitter-jinja2` plugin to the Neovim configuration running tests, and one to provide a simple sample file.

However, the version of this plugin which you forked works neither with Neovim 0.11 nor 0.12, it requires a weird unreleased nightly. Can you please rebase on the current `master`? Then things should work again and I can proceed to record the test results.

tjkeller-xyz:
No problem should be synced up now

HiPhish:
Thanks. We are almost done, I would just like to ask you for one more thing: I want start properly licensing the source files of this project, so I need contributors to add their [REUSE](https://reuse.software/) copyright annotation to the files the submit. In your case that's `rainbow-delimiters.scm`.

Please add a comment like this to the top of the file:

```
;; SPDX-License-Identifier: 0BSD
;; SPDX-FileCopyrightText: © 2026 [your name] <[your contact]>
```

For your name pick whatever you want, it can be your GitHub user name if you don't want to give out your real name.  You can omit the contact part if you want to.

Yeah, I know it's silly for a mostly empty file, but it keeps things clear and auditable.  I want to license all the query files as permissively as possible, so I hope you are OK with my choice.

tjkeller-xyz:
No worries, I went ahead and added it. Thanks!

HiPhish:
Merged, thank you.

PR #196, opened 2025-10-29
Fix ref and label errors for latex; add missing query for C++

sevenninesixthree:


HiPhish:
The C++ changes look good, but I get an "impossible pattern" error for the LaTeX changes. Are you using an older version of the parser?

HiPhish:
Also, for future reference please open separate PRs for different languages. That way if one fix needs work but another one is ready to merge the former does not have to block the latter.

sevenninesixthree:
The original latex is right. I used the old version of parser. Now I delete the changes for latex.

HiPhish:
Merged, thank you.

PR #199, opened 2026-01-11
added htmlangular support

Sckab:
added support for angular templates, including control flow:
```html
@if (cond) {
  ...
}
```
and interpolation:
```html
<span>{{ value }}</span>
```


HiPhish:
Sorry for taking so long. You have called the language `htmlangular`, but according to nvim-treesitter's [SUPPORTED_LANGUAGES](https://github.com/nvim-treesitter/nvim-treesitter/blob/main/SUPPORTED_LANGUAGES.md) document the name is `angular`. Are you running an older version of that plugin? The name of the language is not correct I cannot test it.

Furthermore, the sample file must be named `test/highlight/sample/<language>/<whatever>`, where `language` is the same as the language of the parser and query, and `whatever` is the file name of your choice (usually called `regular`).

Sckab:
oh damn, I named it htmlangular since it's the filetype, when I'll some time I'll fix all the issues!

HiPhish:
Don't worry, I have fixed it myself, I just wanted to double-check if there was a misunderstanding on my side because I don't know Angular. Please check if the patterns are correct (I changed one and added a new one), then we can wrap this up.

asnyder13:
I've checked out that branch and pulled upstream master.  This PR still fixes the plugin for htmlangular filetypes.

HiPhish:
@asnyder13 Thanks for trying out the PR, I have merged it now.

@Sckab I want to clean up the licensing of this project, and I want to license all the query files as `0BSD`. I cannot add the license myself since you are the author, so please add the line `;; SPDX-License-Identifier: 0BSD` to the top of the file and the sample file, or give me permission to do it myself.

Sckab:
yeah you can do it, thanks for apporving the PR!

PR #201, opened 2026-03-06
Add Elixir "do end" block support for colouring

mattrienz:
This adds highlighting support for Elixir "do end" blocks. I also updated the sample test file to cover nested "do end" blocks.

I found this useful locally, and I think it improves readability for nested Elixir blocks.

Since "do end" is a keyword pair rather than punctuation, I realize this may be a judgment call for what should count as a delimiter by default, but I thought it was worth proposing.

### DIFF
--- a/queries/elixir/rainbow-blocks.scm
+++ b/queries/elixir/rainbow-blocks.scm
@@ -1,0 +1,38 @@
+(call
+  (arguments
+    "(" @delimiter
+    ")" @delimiter @sentinel) @container)
+
+(block
+  "(" @delimiter
+  ")" @delimiter @sentinel) @container
+
+(string
+  (interpolation
+    "#{" @delimiter
+    "}" @delimiter @sentinel) @container)
+
+(tuple
+  "{" @delimiter
+  "}" @delimiter @sentinel) @container
+
+(list
+  "[" @delimiter
+  "]" @delimiter @sentinel) @container
+
+(access_call
+  "[" @delimiter
+  "]" @delimiter @sentinel) @container
+
+(bitstring
+  "<<" @delimiter
+  ">>" @delimiter @sentinel) @container
+
+(map
+  "%" @delimiter
+  "{" @delimiter
+  "}" @delimiter @sentinel) @container
+
+(do_block
+  "do" @delimiter
+  "end" @delimiter) @container
--- a/test/highlight/samples/elixir/regular.exs
+++ b/test/highlight/samples/elixir/regular.exs
@@ -28,6 +28,40 @@ defmodule Regular do
     end
 end
 
+defmodule DoEndBlocks do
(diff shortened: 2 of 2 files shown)


HiPhish:
I do like the idea of highlighting matching keywords, and we do already have it in some languages like Lua. However, users could find it too aggressive if entire keywords are highlighted, making the source code look like unicorn puke, so I prefer to provide a different query instead.

Usually this other query is called `rainbow-blocks`. You can see a list of existing alternative queries in `:h rb-delimiters-query`.  Can you please redo this PR such that the `do` and `end` blocks are in a new query called `rainbow-blocks`?

Also, what do you think about adjacent keywords as well? Like `def`, `else` or `rescue`.  The pattern could look something like this (you don't need to `@sentinel` capture):

```
(call
  target: (identifier) @delimiter
  (do_block
    "do" @delimiter
    (else_block
      "else" @delimiter)?
    (rescue_block
      "rescue" @delimiter)?
    "end" @delimiter)) @container
```

I have not actually tried it, so I don't know if it will work. There might be some other keywords to match too, my Elixir is quite rusty.

mattrienz:
Thank you for the reply, I've moved it to the other query.

For adjacent keywords like `def`, `else`, and `rescue`, I think that it’s a judgment call. I would push back on having `def` highlighted as it's usually accompanied with the `do` word that is highlighted. My thought process is to keep this more conservative so it improves readability without becoming too visually aggressive, which could impede readability.

I do see that Lua’s `rainbow-blocks` query already highlights adjacent keywords (such as `if` `else`) , so I’m happy to leave that judgment call to you.

HiPhish:
Let's keep it conservative for the time being. There is also the issue of macros; if a macro is added that used `do` and `end` the query will not work for it, so it would look inconsistent to sometimes have additional keywords highlighted and sometimes not.

Please squash you commits, then I can merge the PR.

mattrienz:
Should be ready to merge.

konnko:
Btw I made a very wild .scm for elixir in [my fork](https://github.com/Konnko/rainbow-delimiters.nvim). Everything that can be theoretically nested gets its color. Not sure if that's even an intended use of rainbow-delimiters so I didn't make a PR(yet? took some time to flesh this out), but if you are interested how you can paint any block — it has it all, works very consistently. 

Here's how it looks in a particularly sloppy written code

image

Also heex queries

HiPhish:
@Konnko Feel free to open a new PR when you are comfortable, then we can take a look. Also, you don't have to maintain an entire extra fork of this plugin if all you want to do is just use your own queries. You can add your own query under whatever name you want to your Neovim configuration and then point the plugin configuration to your query.

```vim
let g:rainbow_delimiters = {'query': {'elixir': 'my-special-rainbows'}}
```

PR #210, opened 2026-04-07
add heex queries, add queries for all elixir blocks

Konnko:
This covers all elixir blocks and heex tags.

However, right now there is a problem with heex when adding any block support to Elixir itself. Specifically, when there is a "cond/case" in <%= %> <% %> tags. They are fundamentally broken at injection because after cleaning up tags we remain with -> with nothing after them which is invalid code that makes Elixir parser to recover and emit wrong ranges.

Visually, it breaks color for all further Elixir code in these tags:
image

Though I should mention colors are somewhat broken even without rainbow-delimiters:
image

I haven't added color to <%= %> <% %> tags because they are pretty rare and this way they would have their own basic color that makes them more visible, just a personal preference.

### DIFF
--- a/queries/elixir/rainbow-blocks.scm
+++ b/queries/elixir/rainbow-blocks.scm
@@ -36,3 +36,112 @@
 (do_block
   "do" @delimiter
   "end" @delimiter) @container
+
+(call
+  target: (identifier) @name @delimiter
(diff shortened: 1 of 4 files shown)


HiPhish:
Looks good, thank you. In the future please open separate PRs for separate languages though, that makes it easier to review things.

PR #212, opened 2026-04-11
Update queries for OCaml

TKristof09:
Update OCaml queries to match recent upstream grammar changes:
  - `let_open_expression` and `let_module_expression` are now part of `let_expression` node
  - `coercion_expression` is now part of `typed_expression` node which already has a separate query.

Relevant tree-sitter-ocaml commits:
https://github.com/tree-sitter/tree-sitter-ocaml/commit/28a93319a67a93780bbd484e6ed5b12dcbe7eddb
https://github.com/tree-sitter/tree-sitter-ocaml/commit/0633bf4be8037ab49743b25e2ab90974c1668aec

Note: As far as I know in OCaml 5.5 many other kinds of `let ... in` expressions will be allowed which will probably need new queries. I'm not yet familiar with the syntax for these since OCaml 5.5 is only in alpha testing as of now so I didn't write the queries for these (even though the tree-sitter grammar seems to have added support for them).

### DIFF
--- a/queries/ocaml/rainbow-blocks.scm
+++ b/queries/ocaml/rainbow-blocks.scm
@@ -12,12 +12,14 @@
     ((let_and_operator) @delimiter _)*)
   "in" @delimiter) @container
 
-(let_open_expression ; Line 268
+(let_expression ; Line 268
   "let" @delimiter
+  (open_module)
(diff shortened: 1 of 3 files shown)


HiPhish:
Done, thank you.

PR #216, opened 2026-07-27
Fix log for local strategy

Thaumy:


HiPhish:
Nice catch, thank you.

PR #217, opened 2026-07-29
Add visible strategy with async streaming delimiter highlighting

Thaumy:
Introduces a new strategy that highlights only the delimiters currently visible in the window.

Rather than highlighting the entire visible region in one blocking pass, the strategy uses future-based async scheduling and streaming match iteration to keep the editor responsive:
- Async futures (future_spawn.lua) — Each highlighting pass is split into small units of work drained in 1ms time slices on a libuv timer, yielding back to the event loop between slices. Futures are processed newest-first (LIFO), so scrolling or editing immediately preempts stale work in favor of the region the user is looking at right now.
- Streaming iteration (hl-node.lua) — A coroutine-driven pre-order DFS walks the syntax tree and yields one match at a time, without buffering anything up front. Each poll call processes a single match, allowing the async scheduler to interleave highlighting with editor input.

A per-buffer cache of highlighted row ranges avoids redundant work on `WinScrolled` (only newly revealed lines are processed). On text change, the cache is invalidated from the first changed row downwards.

Tested on a 10M deeply nested JSON file: the global strategy freezes editor for ~40s, while the visible strategy makes highlighting available immediately — even during fast scrolling.

### DIFF
--- a/lua/rainbow-delimiters/strategy/visible/future_spawn.lua
+++ b/lua/rainbow-delimiters/strategy/visible/future_spawn.lua
@@ -1,0 +1,90 @@
+-- SPDX-License-Identifier: Apache-2.0
+
+-- Cooperative, incremental highlighting.  Highlighting a whole buffer -- or even
+-- just a large visible region -- can involve thousands of extmarks; doing it all
+-- in one go blocks the editor.  Instead a highlighting pass is expressed as a
+-- `poll` function which performs one small unit of work per call and returns
+-- whether any work remains.  Futures are drained one time slice at a time on a
+-- libuv timer, yielding back to the editor in between.
+--
+-- Futures are processed newest-first (the pending futures form a stack).  When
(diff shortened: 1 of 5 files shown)


HiPhish:
Oh wow, this sounds great. If it highlights the entire buffer (or rather its entire visible region), so you think this could replace the global strategy? The lackluster performance of the global strategy has been a thorn in my side for a long time, but I never go around to finding a solution because that would require coming up with a whole new way of doing the highlighting.

I won't have the time to try it out this week yet, but I am excited for a solution to the performance problem.

Thaumy:
This strategy aims to replace the current global default.

Aside from performance improvements, the main user-facing change is how highlighting behaves during heavy operations. For example, when parsing a deeply nested single-line JSON that fills the entire viewport, this new approach applies highlighting incrementally rather than freezing the editor.

I've been using this in my daily work for 2 months and it works well. But feel free to test it longer until you're confident it's stable.

